PR #2, opened 2025-04-03
Onboard vercel AI SDK

rahulkarajgikar:
Adds vercel AI SDK as the default LLM service with support for openai and anthropic.

```
const llmService = createLLMService(config.llm, clientManager, true);
```

Change true above to false to default to other individual LLM services we have defined.

### DIFF
--- a/src/ai/llm/anthropic.ts
+++ b/src/ai/llm/anthropic.ts
@@ -28,22 +28,22 @@ export class AnthropicService implements ILLMService {
         // They don't use a system message like OpenAI,
         // but we can prepend this to the first user message
 
-        const toolDescriptions = Object.entries(tools)
-            .map(([toolName, tool]) => {
-                let description = `- ${toolName}: ${tool.description || 'No description provided'}`;
-                if (tool.parameters && Object.keys(tool.parameters).length > 0) {
-                    description += '\n  Parameters:';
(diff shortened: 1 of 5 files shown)


src/ai/llm/anthropic.ts
```diff
@@ -1,45 +1,49 @@
...
+                        const param = paramRaw as any;
                         description += `\n    - ${paramName}: ${param.description || 'No description'} ${param.type ? `(${param.type})` : ''}`;
                     }
                 }
                 return description;
             })
             .join('\n');
 
-        this.systemContext = `You are Omega, a helpful AI assistant with access to the following tools:\n\n${toolDescriptions}\n\nUse these tools when appropriate to answer user queries. You can use multiple tools in sequence to solve complex problems. After each tool result, determine if you need more information or can provide a final answer.`;
+        this.systemContext = `You are Saiki, a helpful AI assistant with access to the following tools:\n\n${toolDescriptions}\n\nUse these tools when appropriate to answer user queries. You can use multiple tools in sequence to solve complex problems. After each tool result, determine if you need more information or can provide a final answer.`;
```
shaunak99: are these toolDescriptions the same as the in the schema? If so, this is a duplicate since the same data is already appended when given to the 'tools' parameter

rahulkarajgikar: yes good catch, removed the duplicates in latest revision

PR #12, opened 2025-04-05
Change config file format from json to yml 

### LINKED ISSUE #4: Feature: Move config to toml/yaml
toml/yaml has benefits over json
 - more human readable
 - ability to add comments
 - better parsers

rahulkarajgikar:
Yml provides advantages:
 - More human readable
 - Support for comments
 - Support for complex parsing and templates which might come in handy later as the config gets more complex

Updated READMEs to reflect new config 

Resolves #4


app/index.ts
```diff
@@ -33,7 +33,7 @@ if (!existsSync('.env')) {
 program
     .name('mcp-cli')
```
shaunak99: change name to saiki

rahulkarajgikar: done

PR #26, opened 2025-04-07
Update README.md with example

shaunak99:
Added example use case showcasing Saiki for email+slack automation


src/client/mcp-client.ts
```diff
@@ -94,10 +94,17 @@ export class MCPClient implements ToolProvider {
             : `${command} ${args.join(' ')}`;
         logger.info(`Connecting to MCP server: ${serverName}`);
 
+        // Create a properly expanded environment by combining process.env with the provided env
+        const expandedEnv = {
+            ...process.env,
+            ...(env || {})
```
rahulkarajgikar, line 100: for now passing the process.env entirely is fine but maybe a less scoped version is better

rahulkarajgikar (approved):

PR #34, opened 2025-04-09
Add Message management logic into Saiki

### LINKED ISSUE #8: Add unified message management layer
The idea here is to have a standard message management layer that can be used in each LLMService 

Different LLM providers (Anthropic/openai/etc.) have different ways they manage system prompts/message history, so here we would unify this logic for our codebase and provide adapters for each LLM provider. 

This allows us to have standard interfaces and functions that work everywhere and handle the format adaptations internally

rahulkarajgikar:
Resolves #8

## 1. Message Management layer

The goal here is to standardize message/context management across different LLMServices regardless of the formats (openai/anthropic APis expect different messaging formats etc.)

We do this by first creating a new class `MessageManager` as part of each LLMService and delegating all message management to this class.

What does `MessageManager` do:
- Handles adding/deleting messages from the message history
- Has a customizable `formatter` which it delegates to for formatting the message history in the right manner required by that model type (ex: anthropic/openai)
- Has a customizable `tokenizer` which it delegates to for counting tokens (this is important because different LLMs have different ways of counting tokens)
- Has customizable `compressionStrategies` it uses to delegate to for compressing the message history when it gets too large. The priority order of compressionStrategies is also defined by the order in which they are sent into MessageManager.

Keeping tokenizer, formatter, and the compressionStrategies as abstract entities in the MessageManager allows each LLMService to inject specific strategies into their own instance of MessageManager

### Tokenizers

Openai tokenizer - uses `tiktoken` - a public library that approximates openai's token calcualtion
Anthropic and Google - use approximations as of now (4 chars = 1 token). 

There is room for further improvement here.

### Formatters

Openai - formats our internal message history into the message history expected by openai chat completions api
Anthropic - formats our internal message history into the message history expected by anthropic chat  api
Vercel - formats our internal message history into the message history expected by vercel generateText/streamText api.

### Compression Strategies

Initially we just have 2 basic compression strategies

Middle removal - Keep a fixed number of initial messages and last messages, delete everything else
Oldest removal - Keeps removing older messages until we are within the context limit

There is scope for adding more strategies here. Any custom strategy can be injected into the MessageManager while initializing it.

## 2. Folder restructuring

This PR also includes a lot of folder restructuring to accomodate the new features we have added - mostly just moving files around


src/ai/llm/services/openai.ts
rahulkarajgikar, line 1: main diff here is - filepath changed, uses messagemanager now for all message related operations

Issue #16, opened 2025-04-05
Add confirm tool execution as a feature


rahulkarajgikar:
This will allow user to approve tool executions.

Key things needed:
1. A way to intercept any tool request before executing it.
2. Modular system that can be initially simple (yes/no for every execution), but expand to be more complex (remembering preferences)
3. Independent of UI ( should be able to work with CLI inputs (readline) or browser clicks (like Cursor))

Callbacks can help with making the system independent of platform (swap out callbacks to use in UI/CLI, etc.)

*** 
# Proposed initial design

## Overview
Add a confirmation system for tool executions that prompts users before executing tools, making the system more interactive and secure while maintaining flexibility across different environments.

### Core Components

1. **UserConfirmationProvider Interface with Callbacks**
   ```typescript
   interface UserConfirmationProvider {
     requestConfirmation(
       details: ToolExecutionDetails,
       callbacks?: {
         displayDetails?: (details: ToolExecutionDetails) => void;
         collectInput?: () => Promise<string | boolean>;
         parseResponse?: (response: any) => boolean;
       }
     ): Promise<boolean>;
   }
   ```

2. **ToolExecutionDetails Type**
   ```typescript
   interface ToolExecutionDetails {
     toolName: string;
     args: any;
     description?: string;
   }
   ```

3. **Implementation Providers**
   - `CLIConfirmationProvider`: Console-based confirmation using readline
   ```typescript
   export class CLIConfirmationProvider implements UserConfirmationProvider {
     async requestConfirmation(
       details: ToolExecutionDetails,
       callbacks?: ConfirmationCallbacks
     ): Promise<boolean> {
       // Use provided callbacks or defaults
       const display = callbacks?.displayDetails || this.defaultDisplayDetails;
       const collect = callbacks?.collectInput || this.defaultCollectInput;
       const parse = callbacks?.parseResponse || this.defaultParseResponse;
       
       // Display and collect input
       display(details);
       const response = await collect();
       return parse(response);
     }
     
     private defaultDisplayDetails(details: ToolExecutionDetails): void {
       console.log(`Tool: ${details.toolName}`);
       console.log(`Arguments: ${JSON.stringify(details.args, null, 2)}`);
     }
     
     private defaultCollectInput(): Promise<string> {
       // Implementation using readline
       return new Promise(resolve => {
         // readline question implementation
       });
     }
     
     private defaultParseResponse(response: string): boolean {
[...]

PR #42, opened 2025-04-15
Feat/webui

shaunak99:
Added support for WebUI for additional mode of interaction. 

Refer to  CHANGELOG.md for more details.

PR #43, opened 2025-04-15
feat: Tool call confirmation

dieselftw:
## Overview
This PR introduces confirmation for tool calls. I went through the recommended approach in #16 and it made a lot of sense. I've tried to use it as much as possible.

1. Introduced 2 new entities in `ClientManager` class - `allowedTools`, and `confirmationProvider`.
2. This will allow for factoring in user preferences in the future (allow all, allow custom etc) which will be set to `true`. Right now, if no `confirmationProvider` is found, it automatically takes a `DefaultConfirmationProvider` aka everything set to `false`.
3. There are a lot of edge cases/bugs, and formatting that's still needed for this that I'm testing but before I tackle those, I wanted to confirm this is in the direction that you guys would want. Please let me know :)

### 1. Allow tool call
image

### 2. Deny tool call
image

Let me know if you guys think this implementation is what you wanted. If so, I'll complete this PR by this weekend whenever I get time. Thanks!


rahulkarajgikar:
Thanks @dieselftw for raising this PR! 

Will take a look

src/client/confirmation-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+import * as readline from 'readline';
+import chalk from 'chalk';
+import boxen from 'boxen';
+
+/**
+ * Default implementation of UserConfirmationProvider
+ * Automatically approves tools in the allowedTools set,
+ * and prompts for confirmation for other tools using logger's styling
+ */
+export class DefaultConfirmationProvider implements UserConfirmationProvider {
```
rahulkarajgikar: nit: let's rename it to `CLIConfirmationProvider` since readline is specific to CLI and we're adding a web ui in #42

src/client/confirmation-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+            displayDetails?: (details: ToolExecutionDetails) => void;
+            collectInput?: () => Promise<string | boolean>;
+            parseResponse?: (response: any) => boolean;
+        }
+    ): Promise<boolean> {
+        // If the tool is in the allowed list, automatically approve
+        if (this.allowedTools.has(details.toolName)) {
+            logger.debug(`Tool '${details.toolName}' is pre-approved for execution`);
+            return true;
+        }
```
rahulkarajgikar: Nice! Useful re-usable implementation.

Another thing I was thinking is that users might want to pre-approve all tools in certain MCP servers to avoid having to approve each time they re-run Saiki

So maybe it could make sense to add a config parameter in the MCP server config  [`SSEServerConfig` and `StdioServerConfig`] called `toolsRequireApproval` - set to true by default (to keep user behaviour the same)

If `toolsRequireApproval` = true for an MCP server, then we automatically add all the tools from there as allowedTools in the `ClientManager`

If `toolsRequireApproval` = false for the MCP server, then follow this flow of requiring approval

this way we let users decide if they want to have tool confirmations or not - what do you think?

src/client/manager.ts
```diff
@@ -60,6 +69,33 @@ export class ClientManager {
...
+    }
+
+    /**
+     * Allow a tool to execute without confirmation
+     * @param toolName Name of the tool to allow
+     */
+    allowTool(toolName: string): void {
+        this.allowedTools.add(toolName);
+        logger.info(`Tool '${toolName}' has been allowed for execution without confirmation`);
+    }
```
rahulkarajgikar: maybe instead of maintaining `allowedTools` in both ClientManager and the ConfirmationProvider, we keep `allowedTools` in the confirmation provider and use this method to delegate to the confirmation provider.

example pseudo code:
```
 allowTool(toolName: string): void {
        this.confirmationProvider.allowTool(toolName);
        logger.info(`Tool '${toolName}' has been allowed for execution without confirmation`);
    }
```

this way we let the confirmation provider deal with maintaining the state of allowed tools and also approving the tools.

similar for `disallowTool`  and `isToolAllowed`

let me know what you think. open to keeping it here as well if we have good reasons to

src/client/confirmation-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+
+        // Display tool call using the logger's built-in method
+        logger.toolCall(details.toolName, details.args);
+
+        // Collect user input with arrow key navigation
+        return await this.collectArrowKeyInput();
+    }
+
+    /**
+     * !!!! This is an AI-generated sloppy implementation, TODO: REFACTOR THIS !!!
```
rahulkarajgikar: haha been there

src/client/manager.ts
```diff
@@ -1,12 +1,20 @@
...
 export class ClientManager {
     private clients: Map<string, ToolProvider> = new Map();
     private connectionErrors: { [key: string]: string } = {};
     private toolToClientMap: Map<string, ToolProvider> = new Map();
+    private allowedTools: Set<string> = new Set();
+    private confirmationProvider: UserConfirmationProvider;
+
+    constructor(confirmationProvider?: UserConfirmationProvider) {
+        // If a confirmation provider is passed, use it, otherwise use the default implementation
+        this.confirmationProvider = confirmationProvider ?? new DefaultConfirmationProvider(this.allowedTools);
```
rahulkarajgikar: i'm thinking we don't use a default here and keep `confirmationProvider` as a mandatory field.

then we let the initializer of `ClientManager` decide which `ConfirmationProvider` to use

this would tie in well with this change in #42: https://github.com/truffle-ai/saiki/pull/42/files#diff-b531b96dd0079473aedc9e22fe110d7e1f5bac39a62a788614227b710811fe94

`index.ts` would be the central routing point. we could add another parameter called `displayInterface` to `initializeServices` and then `initializeServices` would initialize the `ClientManager` with the right confirmation provider based on if the runMode is web or cli.

this might be a bit complicated haha but let me know if this makes sense

src/client/confirmation-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+     * @param callbacks Optional callbacks for customizing the confirmation flow
+     * @returns Promise resolving to boolean indicating if execution is approved
+     */
+    async requestConfirmation(
+        details: ToolExecutionDetails,
+        callbacks?: {
+            displayDetails?: (details: ToolExecutionDetails) => void;
+            collectInput?: () => Promise<string | boolean>;
+            parseResponse?: (response: any) => boolean;
+        }
```
rahulkarajgikar: i know we haven't used the callbacks yet since this is just an initial implementation but yeah wanted to mention some points here before we do this part.

We've actually started using node's `EventEmitter` class to handle events (check the different LLMServices). This leads to generally cleaner implementations - the class creating the events just needs to focus on emitting them, and the class who is interested in the events just needs to subscribe to the events.

Event based pub-sub might be a better choice than callbacks since we just need to register subscribers to the event bus instead of passing callbacks all the way to the start point. 

Check out `VercelLLMService`, `AgentEventManager` classes to understand this a bit better.

Just something to think about here before we commit to an implementation.

rahulkarajgikar (changes requested):
@dieselftw thanks for the PR! 

Have added some points to think about and a few comments to address but overall the implementation looks really solid!

These are the points though:
1. Giving users a way to configure if they want to use this approval flow as part of the MCP server config is useful and avoids redundant approvals every time they restart Saiki
2. Few minor implementation changes to make the code more interface independent based on WebUI/CLI.
3. Might make sense to let the ConfirmationProvider handle maintaining the state of allowedTools
4. Choice between callbacks/event based mechanism for propagating info back to the UI/CLI for modals/popup - something to think about

Happy to hear your thoughts and discuss more!

rahulkarajgikar:
fyi just merged #42 adding the webUI

rezerov:
Thanks for the in-depth review! Will get to this soon.

src/client/confirmation-provider.ts
rezerov: That's one way to do it. Although, since we are integrating it with a WebUI, I think a settings/preferences tab will be present there for authenticated users. If there's a settings tab which we can pull the values of in code via `getUserSettings()`, it might make sense to have this flag as part of `UserConfirmationProvider`. 

This would allow further granular control over what kinds of permission you want to grant the user (pass an array of allowed MCP servers or individual tools, if not present, ask for confirmation), and another `requireApproval` flag as you mentioned.

Your suggestion also seems valid, but tool approval settings built into the server config kinda adds unnecessary coupling imo. What do you think?

src/client/manager.ts
rezerov: Good catch. Fixed :)

src/client/confirmation-provider.ts
rahulkarajgikar: Agree that keeping it in the server config isn't ideal. The main reason i wanted it in the server config was so that it could persist across multiple runs of saiki - bad UX if someone needs to keep resetting this field every time they want to run saiki

One advantage of keeping it as part of the server config is that we get granular control for each server. 
But now that i think more about it, there are probably better ways to handle this.

Having a Settings/UserSettings class could probably be a better way to maintain users state. Singleton class is a good contender so that any class can consume the same global settings easily

Maybe we introduce a separate file (or separate section in the existing config file) for user preferences and use that to persist user settings? That could work along with your idea of `getUserSettings` - which could just read these settings from the file. 
we could also have a `changeUserSetting` or `updateUserSettings` function which writes to the settings file. 

How this could work in the web UI:
 - expose APIs that talk to `changeUserSetting` and `updateUserSettings` to allow users to tune the settings

How this could work in the CLI:
- have hotkeys for changing settings or let user directly update the settings file. claude code does it in an interesting way - each time you run a new tool and approve it, it remembers that.

What do you think? We've merged the webUI code as well btw so you can take a look at that and see if this makes sense.

rahulkarajgikar: Ok i've given this some more thought and here's what i've come up with. 

The main requirement here is that we need a way to persist settings somehow for a good UX. The easiest way we can do that is through a file, something like `~/.saiki/settings.json`. 

But we also don't want to tie down to one specific way of storing settings - there might be other requirements based on where saiki is running.

Let's say we first create a new `SettingsProvider` interface

```
export interface SettingsProvider {
  getUserSettings(userId?: string): Promise<UserSettings>;
  updateUserSettings(userId?: string, settings: UserSettings): Promise<void>;
  // ...etc
}
```

we can then create a FileSettingsProvider that uses `~/.saiki/settings.json` for reading/writing settings:

```
export class FileSettingsProvider implements SettingsProvider {
  // Ignores userId, always uses ~/.saiki/settings.json
}
```

then in `ConfirmationProvider`, instead of storing the allowed tools or preferences directly in this class, we delegate to `SettingsProvider` to read/write the settings :

something like this, the below is just pseudocode to give you an idea, but see here how we delegate the fetching/checking of the current state of the settings to the settings provider. 
```
export class ConfirmationProvider {
  constructor(private settingsProvider: SettingsProvider) {}

  async confirmTool(server: string, tool: string, userId?: string): Promise<boolean> {
    if (await this.settingsProvider.isToolApproved(server, tool, userId)) {
      return true;
    }
    // Prompt user for approval (CLI: prompt, Web: modal)
    // If approved:
    await this.settingsProvider.approveTool(server, tool, userId);
    return true;
  }
}
```

This approach gives us flexibility - say later on we want to store user settings in a db instead of a file, we would then:

1. Implement a new `DBSettingsProvider` like this:
```
export class DbSettingsProvider implements SettingsProvider {
  async getUserSettings(userId: string): Promise<UserSettings> {
    // Query DB for userId
  }
  // ...other methods, always use userId
}
```

2. Update ConfirmationProvider to use the `DBSettingsProvider` instead of the `FIleSettingsProvider`
example: this could even be controlled by an env variable:

```
 const settingsProvider = process.env.SAIKI_USE_DB_SETTINGS
    ? new DbSettingsProvider()
    : new FileSettingsProvider();
```


To begin with we can skip the `DbSettingsProvider` and just use a common `FileSettingsProvider` in both the webUI and CLI - this is because we want to keep the webUI and CLI experience identical and aren't expecting multiple users. But this approach allows for future extensibility later (in case we want a multi-tenant agent to deal with settings from different users)

Let me know what you think of this approach

rezerov: Sounds good.

rezerov: Implemented an initial version of this. Let me know what you think of it. This is a very fundamental change to the repo. If I went wrong somewhere, please let me know.

src/client/manager.ts
rezerov: Interesting. This is indeed a bit complicated. Do you want to take this as an argument for `ClientManager`? Like how will the initializer of client manager pass the information to `ClientManager`

rezerov:
@rahulkarajgikar I've made a lot of changes, including a settings provider. Please check it out and let me know if it's to your liking. I haven't formatted the code properly yet, will do that once we're done with implementation. As of now, this implementation does work (try switching `toolApprovalRequired` to false in `src/providers/settings/settings.ts`)

src/client/providers/settings/settings.ts
```diff
@@ -0,0 +1,8 @@
+import { UserSettings } from "../../types.js";
+
+export const settings: UserSettings = {
+    userId: 1,
+    settings: {
+        toolApprovalRequired: true,
+    },
+}
```
rahulkarajgikar: let's rename this file to `src/settings/utils.ts` and fetch these settings from a function `getDefaultUserSettings`

src/client/providers/confirmation-provider.ts
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Default implementation of UserConfirmationProvider
+ * Automatically approves tools in the allowedTools set,
+ * and prompts for confirmation for other tools using logger's styling
+ */
+export class CLIConfirmationProvider implements UserConfirmationProvider {
+    public allowedTools: Set<string>;
+    private rl: readline.Interface | null = null;
+    public settings: Promise<UserSettings>;
```
rahulkarajgikar: instead of storing settings, store the `settingsProvider` and use that in this class to fetch settings - this let's the `settingsProvider` handle reading/writing all the settings

this will become handy later on if settings have more features later - we just need to add more methods into `settingsProvider`

src/client/providers/confirmation-provider.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    constructor(allowedTools?: Set<string>, settingsProvider?: SettingsProvider) {
+        this.allowedTools = allowedTools ?? new Set();
+        this.settings = new FileSettingsProvider().getUserSettings();
```
rahulkarajgikar: this line would then change to be

```
this.settingsProvider = settingsProvider ?? new FileSettingsProvider()
```

and then later places, wherever we check for settings

src/client/providers/settings/settings.ts
```diff
@@ -0,0 +1,8 @@
+import { UserSettings } from "../../types.js";
+
+export const settings: UserSettings = {
+    userId: 1,
+    settings: {
+        toolApprovalRequired: true,
```
rahulkarajgikar: let's set this to default false until we figure out how to store the settings in a file

this way we can merge the PR soon and keep the scope of the PR small

src/client/manager.ts
```diff
@@ -131,6 +155,14 @@ export class ClientManager {
         return this.connectionErrors;
     }
 
+    /**
+     * Get all allowed tools
+     * @returns Set of tool names that are allowed without confirmation
+     */
+    getAllowedTools(): Set<string> {
+        return new Set(this.confirmationProvider.allowedTools);
```
rahulkarajgikar: don't need to create a new set, can just do:
```
return this.confirmationProvider.allowedTools
```

src/client/providers/file-settings-provider.ts
```diff
@@ -0,0 +1,14 @@
+import { SettingsProvider, UserSettings } from "../types.js";
+import { settings } from "./settings/settings.js";
+
+export class FileSettingsProvider implements SettingsProvider {
+    private settings: UserSettings = settings;
```
rahulkarajgikar: rename this to `userSettings` to avoid `settings.settings`

rahulkarajgikar (changes requested):
Nice job on the working tool execution implementation!

Firstly on the file structure, I was thinking something like this - this way we keep settings and confirmation separate.

```
src/
├── settings/
│   ├── types.ts
│   └── file-settings-provider.ts
└── client/
    └── tool-confirmation/
        ├── types.ts
        └── cli-confirmation.ts
```

Also yeah, the scope of the PR is getting quite big and i want to limit this.

So for now, I'm thinking we make these changes:
1. file structure changes i've mentioned
2. comments i've mentioned
3. changing toolApproval to be `false` by default - until we figure out more things like storing into the file etc.
```

and merge the PR. this way we get some progress and have a new opt-in feature and keep the PR scope small


remaining changes like the file system implementation, event emission we can take incrementally and flip default `false` to `true` later on

what do you think?

rezerov:
sounds good! will make the changes soon

rezerov:
All changes done @rahulkarajgikar

rahulkarajgikar (approved):

rahulkarajgikar:
Thanks for raising this PR @dieselftw! I've merged it :)

PR #57, opened 2025-04-17
Feature support for connecting to server via WebUI 

shaunak99:


src/client/manager.ts
```diff
@@ -115,6 +116,33 @@ export class ClientManager {
...
     }
 
+    /**
+     * Dynamically connect to a new MCP server.
+     * @param name The unique name for the new server connection.
+     * @param config The configuration for the server.
+     * @returns Promise resolving when the connection attempt is complete.
+     * @throws Error if the connection fails.
+     */
+    async connectClient(name: string, config: McpServerConfig): Promise<void> {
```
rahulkarajgikar, line 138: we have the same code in https://github.com/truffle-ai/saiki/pull/57/files#diff-c2043b551fcabbf98e56573bf9459c21925e67d5a526087cf71ef5ab5636819fR93-R101 - we can update ` initializeFromConfig` to use this new function `connectClient` to avoid duplicate code

app/web/server.ts
rahulkarajgikar, line 1: formatting doesn't look right, can you run `npm run format`

change vscode/cursor settings in bottom right to use 4 spaces and it should match formatting

app/web/client/script.js
```diff
@@ -1,208 +1,397 @@
...
+                    resultString = JSON.stringify(message.data.result, null, 2);
+                } catch {
+                    resultString = String(message.data.result);
+                }
+                const headerHtml = `<strong>Tool Result:</strong> ${escapeHtml(message.data.toolName)}`;
+                const contentHtml = `<pre><code>${escapeHtml(resultString)}</code></pre>`;
+                appendExpandableMessage(headerHtml, contentHtml, 'tool-result');
+                currentAiMessageElement = null;
+                break;
+            }
```
rahulkarajgikar, line 202: toolResult isn't showing up in the UI

either toolResult isn't reaching till here or there is a bug in these lines of code

update: looks like this isn't showing in CLI either will investigate this separately

app/web/client/script.js
```diff
@@ -213,94 +402,6 @@ function escapeHtml(unsafe) {
...
-// Initial setup
-messageInput.disabled = true; // Disable input until connected
-sendButton.disabled = true;
-resetButton.disabled = true;
-connectWebSocket(); // Start WebSocket connection
-adjustTextareaHeight(); // Adjust initial height 
\ No newline at end of file
+// Note: isScrolledToBottom was moved inside as it directly uses messageListWrapper fetched there.
+// Other WebSocket logic (connectWebSocket, handleWebSocketMessage) is defined inside DOMContentLoaded
+// as it needs access to the helper functions defined there.
```
rahulkarajgikar: don't need this

app/web/server.ts
```diff
@@ -47,28 +47,89 @@ export async function initializeWebUI(
...
+    }
+  });
+  // ---- END NEW ENDPOINT ----
+
+  
+  // ---- NEW ENDPOINT: Connect to a new MCP Server ----
+  app.post('/api/connect-server', express.json(), async (req, res) => {
+    logger.info('Received request via POST /api/connect-server');
+    const { name, config } = req.body;
+
```
rahulkarajgikar: this is the exact same as the previous endpoint so we have duplicate code for `/api/connect-server`, let's delete

rahulkarajgikar (changes requested):

app/web/server.ts
```diff
@@ -47,28 +47,89 @@ export async function initializeWebUI(
...
+      const errorMessage = error instanceof Error ? error.message : 'Unknown error during connection';
+      logger.error(`Error handling POST /api/connect-server for '${name}': ${errorMessage}`);
+      // Send back a more specific error if possible (e.g., distinguish between bad config and connection failure)
+      res.status(500).send({ error: `Failed to connect to server '${name}': ${errorMessage}` });
+    }
+  });
+  // ---- END NEW ENDPOINT ----
+
+  
+  // ---- NEW ENDPOINT: Connect to a new MCP Server ----
```
rahulkarajgikar: don't need stuff like 'NEW ENDPOINT" because it will not remain new for long. can remove this for any endpoints added

app/web/client/script.js
rahulkarajgikar: it was an issue with vercel llm service when the tool gets called in first agent loop

fixed in https://github.com/truffle-ai/saiki/pull/59

rahulkarajgikar (approved):

PR #62, opened 2025-04-18
implement system prompt contributors to break up system prompt logic

rahulkarajgikar:
-- this is just a draft PR. This pr description was AI generated --

## Summary
- Implement system prompt contributors to have granular control over system prompt
- Update LLM service integrations
- Update .gitignore to exclude local/dev files

---
This PR was prepared after completing all tracked work in the local fork (rahulkarajgikar/saiki) and is ready for review and merge into truffle-ai/saiki.


src/ai/llm/messages/manager.ts
```diff
@@ -214,54 +233,11 @@ export class MessageManager {
...
-        return this.systemPrompt;
-    }
-
-    /**
-     * Counts total tokens in current conversation including system prompt.
-     * Requires a tokenizer to be configured.
-     * @returns Total token count or null if no tokenizer is available.
-     */
-    countTotalTokens(): number | null {
-        if (!this.tokenizer) return null;
```
rahulkarajgikar: this function was incorrectly deleted. we need to add it back

src/ai/llm/messages/manager.ts
```diff
@@ -295,10 +271,10 @@ export class MessageManager {
      * @returns Formatted system prompt or null/undefined based on formatter implementation
      * @throws Error if formatting fails
      */
-    getFormattedSystemPrompt(): string | null | undefined {
+    async getFormattedSystemPrompt(): Promise<string | null | undefined> {
+        const prompt = await this.getSystemPrompt();
         try {
-            // Check if the formatter implements getSystemPrompt and call it
-            return this.formatter.getSystemPrompt?.(this.systemPrompt);
+            return this.formatter.getSystemPrompt?.(prompt ?? undefined);
```
rahulkarajgikar: unrelated but we could rename `getSystemPrompt` in `this.formatter` to `formatSystemPrompt` for clarity

src/ai/llm/messages/manager.ts
rahulkarajgikar: nvm it was added back later in the class

src/ai/llm/services/anthropic.ts
```diff
@@ -18,6 +21,8 @@ export class AnthropicService implements ILLMService {
     private clientManager: ClientManager;
     private messageManager: MessageManager;
     private eventEmitter: EventEmitter;
+    private systemPromptBuilder?: SystemPromptBuilder;
+    private promptContext?: PromptContext;
```
rahulkarajgikar: now both the llmservice and the messagemanager have systemPromptBuilder and promptContext - maybe there is a better way of doing this? would like to know thoughts

src/ai/llm/services/anthropic.ts
```diff
@@ -81,8 +109,15 @@ export class AnthropicService implements ILLMService {
...
-                const systemPrompt = this.messageManager.getFormattedSystemPrompt();
+                let messages = this.messageManager.getFormattedMessages();
+                // Ensure all message content fields are arrays of content blocks
+                messages = messages.map((msg) => {
+                    if (typeof msg.content === 'string') {
+                        return { ...msg, content: [{ type: 'text', text: msg.content }] };
+                    }
+                    return msg;
+                });
+                const systemPrompt = await this.messageManager.getFormattedSystemPrompt();
```
rahulkarajgikar, line 94: do we really need this check for message content fields being arrays of content blocks?

src/ai/llm/services/vercel.ts
```diff
@@ -56,6 +74,9 @@ export class VercelLLMService implements ILLMService {
         return this.clientManager.getAllTools();
     }
 
+    /**
+     * @deprecated Use setSystemPromptBuilder instead for modular system prompt support.
+     */
```
rahulkarajgikar: we can just delete unused functions instead of marking them as deprecated

src/ai/llm/services/types.ts
```diff
@@ -9,9 +11,17 @@ export interface ILLMService {
...
     completeTask(userInput: string): Promise<string>;
 
-    // Update the system message/context
+    /**
+     * @deprecated Use setSystemPromptBuilder instead for modular system prompt support.
+     */
     updateSystemContext(newSystemPrompt: string): void;
 
+    // Accept a SystemPromptBuilder instance
+    setSystemPromptBuilder(builder: SystemPromptBuilder): void;
```
rahulkarajgikar: i'm noticing we added this in both `ILLMService` and `MessageManager`. Is this intentional? seems like duplicate things and we now need to handle passing the objects around from LLMService to messagemanager

src/ai/systemPrompt/SystemPromptRegistry.ts
```diff
@@ -0,0 +1,57 @@
+import { ISystemPromptContributor } from './types.js';
+import { SystemPromptBuilder } from './SystemPromptBuilder.js';
+
+/**
+ * Singleton registry for system prompt contributors.
+ */
+export class SystemPromptRegistry {
```
rahulkarajgikar: where are we using this class? i don't see it being used anywhere

rahulkarajgikar (commented):
reviewed AI generated PR that saiki generated!

src/ai/llm/messages/manager.ts
```diff
@@ -201,67 +210,22 @@ export class MessageManager {
...
-                total += overheadPerMessage;
-            }
-        } catch (error) {
-            logger.error('MessageManager: Error counting tokens:', error);
-            // Decide on error handling: return null, throw, or return current total?
-            // Returning null indicates counting failed.
-            return null;
+    async getSystemPrompt(): Promise<string> {
+        if (!this.contributors.length) {
+            throw new Error('MessageManager: No contributors found.');
```
rahulkarajgikar: Change the error to `MessageMnager: No system prompt contributors found while generating system prompt` to make it more descriptive

src/ai/llm/messages/manager.ts
```diff
@@ -279,7 +243,7 @@ export class MessageManager {
 
         try {
             // Pass a read-only view of the potentially compressed history to the formatter
-            return this.formatter.format([...this.history], this.systemPrompt);
+            return this.formatter.format([...this.history], undefined); // systemPrompt is now handled by getSystemPrompt
```
rahulkarajgikar: passing `undefined` here is weird. if we don't need to pass systemPrompt here, we should just remove this argument completely and update comments in the formatter `format` interface and concrete implementations docs.

src/ai/llm/messages/manager.ts
```diff
@@ -201,67 +210,22 @@ export class MessageManager {
...
-                }
-                // Add overhead for each message structure
-                total += overheadPerMessage;
-            }
-        } catch (error) {
-            logger.error('MessageManager: Error counting tokens:', error);
-            // Decide on error handling: return null, throw, or return current total?
-            // Returning null indicates counting failed.
-            return null;
+    async getSystemPrompt(): Promise<string> {
```
rahulkarajgikar, line 217: let's add a debug log at the beginning and ending of this function for easier debugging

src/ai/llm/messages/manager.ts
```diff
@@ -394,4 +358,47 @@ export class MessageManager {
...
+     */
+    countTotalTokens(): number | null {
+        if (!this.tokenizer) return null;
+
+        let total = 0;
+        const overheadPerMessage = 4; // Approximation for message format overhead
+
+        try {
+            // Count system prompt from cache
+            if (this.systemPrompt) {
```
rahulkarajgikar, line 407: add a debug log here something like: 

logger.debug(`[MessageManager]: Using cached system prompt: <print the systemprompt> while counting tokens`

src/ai/systemPrompt/contributors.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(id: string, priority: number, content: string) {
+    if (!content) {
+      throw new Error(`StaticContributor '${id}' requires non-empty content.`);
+    }
+    this.id = id;
+    this.priority = priority;
+    this.staticContent = content;
+  }
+
+  async getContent(_context: DynamicContributorContext): Promise<string> {
```
rahulkarajgikar, line 20: a little weird for this function to accept a `DynamicContributorContext` - doesn't fully make sense for `StaticContributor` to accept a `DynamicContributorContext`, even if we are skipping it

src/ai/systemPrompt/loader.ts
rahulkarajgikar, line 1: we earlier had different frequencies of updating the system prompt for each contributor - per 'session', per 'message', or at a defined `timeInterval`. what would it take to have something similar here?

src/ai/llm/messages/manager.ts
rahulkarajgikar: edit: this doesn't seem right. now i see that the system prompt isn't even going into the messages in VercelLLMService

2 ways i can think of to solve this:

1. add an argument 'includeSystemPrompt` to this function and pass systemPrompt if needed here. in anthropicservice, don't pass this arg but in the other services do pass it.
2. or in `VercelLLMService` (and potentially others), enforce that we call `getSystemPrompt` first

rahulkarajgikar (commented):
added commentts on new implementation

rahulkarajgikar:
closed in #80

PR #63, opened 2025-04-18
Fix/puppeteer

shaunak99:
- Upgraded the puppeteer server.
- Upgraded saiki.yml
- Temporary patch in openai tokenizer to support gpt- 4.1 series


shaunak99:
amazon_shopping_demo

src/ai/llm/tokenizer/openai.ts
```diff
@@ -16,7 +16,7 @@ export class OpenAITokenizer implements ITokenizer {
     constructor(model: string) {
         try {
             // Ensure the model is a valid TiktokenModel type
-            this.model = model as TiktokenModel;
+            this.model = (model.includes('gpt-4.1-') ? model.replace('gpt-4.1-', 'gpt-4o-') : model) as TiktokenModel;
```
rahulkarajgikar: let's move this to a function `fetchTikTokenModel()` - it extends better if we need to have similar conversions for more openai models.

```
this.model = fetchTiktokenModel(model)
```

store the unsupported models and the default config as constants at the top of the file
```
UNSUPPORTED_MODELS[] = ['gpt-4.1',.....]
DEFAULT_MODEL = 'gpt-4o'
```

do the conversion here:
```
fetchTiktokenModel(model):
    // first check if model is unsupported - if so, return default model
    for entry in unsupportedModels, if model.includes(entry) - return DEFAULT_MODEL
    
    // if we get here, it means model is supported and we can return it back)
    return model
```

package.json
```diff
@@ -52,7 +52,10 @@
         "marked": "^15.0.8",
         "openai": "^4.89.0",
         "ora": "^7.0.1",
+        "playwright": "^1.52.0",
         "puppeteer-core": "^24.4.0",
+        "puppeteer-extra": "^3.3.6",
+        "puppeteer-extra-plugin-stealth": "^2.11.2",
```
rahulkarajgikar: doesn't look like we're using these anywhere, we can uninstall to keep our package image small

README.md
```diff
@@ -99,8 +99,26 @@ Check out [example configurations](./configuration/examples/) or the [Configurat
 
 ## 🎯 Examples
 
+### 🛒 Amazon Shopping Assistant
```
rahulkarajgikar, line 102: nice!

rahulkarajgikar (changes requested):

src/ai/llm/tokenizer/openai.ts
rahulkarajgikar: this way if we have another model later that is unsupported, we just need to modify `UNSUPPORTED_MODELS`

package.json
shaunak99: Done!

src/ai/llm/tokenizer/openai.ts
shaunak99: Hey, updated the OpenAI tokenizer based on your feedback:

1.  Pulled the model checking stuff into its own function (`fetchTiktokenModel`). It handles the weird model prefixes and tries to make sure tiktoken knows the model, falling back to the default (`gpt-4o`) if not.
2.  Made `UNSUPPORTED_PREFIXES` and `DEFAULT_MODEL` constants at the top so they're easy to find and change.
3.  The main `OpenAITokenizer` class now just calls that new function to figure out the right model name.

src/ai/llm/tokenizer/openai.ts
```diff
@@ -1,6 +1,34 @@
...
+    for (const prefix of UNSUPPORTED_PREFIXES) {
+        if (model.startsWith(prefix)) {
+            console.warn(`Model '${model}' might not be directly supported by tiktoken. Using default '${DEFAULT_MODEL}'.`);
+            return DEFAULT_MODEL;
+        }
+    }
+
+    try {
+        // Verify the model name is known to tiktoken
+        encoding_for_model(model as TiktokenModel);
```
rahulkarajgikar: remove this `encoding_for_model`, we are doing this later. we can just return the model

and we don't need a try catch block here either

shaunak99: updated

rahulkarajgikar (approved):

Issue #38, opened 2025-04-10
Add multimodal support


rahulkarajgikar:
Add support for image/file/video based inputs

shaunak99:
Added support for image input

PR #64, opened 2025-04-21
Add support for image parts/inputs

shaunak99:
Updated the webui, Message Manager and vercel services for supporting image inputs.

Additionally, response emission in vercel service was moved to the onStepFunction to also capture thinking tokens generated along side toolCalls.

Vercel service was updated to add toolCalls and responses to message manager to maintain context throughout chat session.

Please review current implementation and suggest improvements. After review and edits, will also apply the changes to openai and anthropic services,

### DIFF
--- a/app/web/client/script.js
+++ b/app/web/client/script.js
@@ -12,7 +12,6 @@ marked.use({ sanitize: true });
 // --- Wait for DOMContentLoaded before initializing UI and listeners ---
 document.addEventListener('DOMContentLoaded', () => {
     // --- Get DOM Elements --- 
-    const messageList = document.getElementById('message-list');
     const messageListWrapper = document.getElementById('message-list-wrapper');
     const messageInput = document.getElementById('message-input');
     const sendButton = document.getElementById('send-button');
@@ -33,7 +32,7 @@ document.addEventListener('DOMContentLoaded', () => {
     const removeImageBtn = document.getElementById('remove-image-btn');
 
     // --- Check if critical elements exist ---
-    if (!messageList || !messageListWrapper || !userInput || !sendButton || !resetButton || !statusIndicator || !connectServerButton || !modal || !connectServerForm || !serverTypeSelect || !stdioOptionsDiv || !sseOptionsDiv || !chatLog || !imageUpload || !imagePreviewContainer || !imagePreview || !removeImageBtn) {
+    if (!chatLog || !messageListWrapper || !userInput || !sendButton || !resetButton || !statusIndicator || !connectServerButton || !modal || !connectServerForm || !serverTypeSelect || !stdioOptionsDiv || !sseOptionsDiv || !imageUpload || !imagePreviewContainer || !imagePreview || !removeImageBtn) {
         console.error("Initialization failed: One or more required DOM elements not found.");
         // Display error to user if possible
         if (chatLog) {
@@ -43,19 +42,6 @@ document.addEventListener('DOMContentLoaded', () => {
              chatLog.appendChild(errorElement);
         }
         return; // Stop further execution
-    }
-
-    // --- Helper Functions (Defined inside DOMContentLoaded) ---
-
-    // Function to update the status indicator
-    function updateStatus(text, className) {
-        if (statusIndicator) { // Check if element exists
-           statusIndicator.className = className; // Apply class for styling (e.g., connected, error, thinking)
-           statusIndicator.setAttribute('data-tooltip', text); // Update tooltip text
-           console.log(`Status updated: ${text} (${className})`); // Optional logging
-        } else {
-           console.error("Status indicator element not found.");
-        }
     }
 
     function scrollToBottom() {
@@ -120,7 +106,7 @@ document.addEventListener('DOMContentLoaded', () => {
             thinkingIndicatorElement.appendChild(innerSpan);
             thinkingIndicatorElement.setAttribute('role', 'status');
             thinkingIndicatorElement.setAttribute('aria-live', 'polite');
-            messageList.appendChild(thinkingIndicatorElement);
+            chatLog.appendChild(thinkingIndicatorElement);
             if (shouldScroll) {
                 scrollToBottom();
             }
@@ -169,9 +155,6 @@ document.addEventListener('DOMContentLoaded', () => {
 
         // Send via WebSocket instead of fetch
         if (ws && ws.readyState === WebSocket.OPEN) {
-            // Indicate processing state without clearing 'connected'
-            statusIndicator.classList.add('thinking');
-            statusIndicator.setAttribute('data-tooltip', 'Processing...');
             const payload = { type: 'message', content: messageToSend, imageData: imageDataToSend };
             ws.send(JSON.stringify(payload));
         } else {
@@ -256,9 +239,7 @@ document.addEventListener('DOMContentLoaded', () => {
                     messageElement.innerHTML = finalHtmlContent; // Set the parsed HTML content
                     chatLog.appendChild(messageElement);
                 }
-                // remove thinking indicator only
-                statusIndicator.classList.remove('thinking');
-                statusIndicator.setAttribute('data-tooltip', 'Connected');
+                // done handling AI response
                 break;
             }
             case 'toolCall': {
@@ -288,7 +269,7 @@ document.addEventListener('DOMContentLoaded', () => {
                 break;
             }
             case 'conversationReset': {
-                messageList.innerHTML = '';
+                chatLog.innerHTML = '';
                 displaySystemMessage('Conversation history cleared.', 'status');
                 currentAiMessageElement = null;
                 shouldScroll = true;
@@ -338,7 +319,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
         ws.onerror = (error) => {
             console.error('WebSocket error:', error);
-            statusIndicator.classList.remove('connected', 'thinking');
+            statusIndicator.classList.remove('connected');
             statusIndicator.classList.add('error');
             statusIndicator.setAttribute('data-tooltip', 'Connection Error');
             displaySystemMessage('WebSocket connection error. Please try refreshing.', 'error');
@@ -349,7 +330,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
         ws.onclose = (event) => {
             console.log(`WebSocket connection closed. Code: ${event.code}, Reason: ${event.reason}`);
-            statusIndicator.classList.remove('connected', 'thinking');
+            statusIndicator.classList.remove('connected');
             statusIndicator.classList.add('error');
             statusIndicator.setAttribute('data-tooltip', `Disconnected: ${event.reason || 'Connection closed'}`);
             messageInput.disabled = true;
--- a/app/web/client/script.js
+++ b/app/web/client/script.js
@@ -12,7 +12,7 @@ marked.use({ sanitize: true });
 // --- Wait for DOMContentLoaded before initializing UI and listeners ---
 document.addEventListener('DOMContentLoaded', () => {
     // --- Get DOM Elements --- 
-    const messageListWrapper = document.getElementById('message-list-wrapper');
+    const chatLog = document.getElementById('message-list');
     const messageInput = document.getElementById('message-input');
     const sendButton = document.getElementById('send-button');
     const resetButton = document.getElementById('reset-button');
@@ -24,15 +24,13 @@ document.addEventListener('DOMContentLoaded', () => {
     const serverTypeSelect = document.getElementById('server-type');
     const stdioOptionsDiv = document.getElementById('stdio-options');
     const sseOptionsDiv = document.getElementById('sse-options');
-    const chatLog = document.getElementById('message-list');
-    const userInput = document.getElementById('message-input');
     const imageUpload = document.getElementById('image-upload');
     const imagePreviewContainer = document.getElementById('image-preview-container');
     const imagePreview = document.getElementById('image-preview');
     const removeImageBtn = document.getElementById('remove-image-btn');
 
     // --- Check if critical elements exist ---
-    if (!chatLog || !messageListWrapper || !userInput || !sendButton || !resetButton || !statusIndicator || !connectServerButton || !modal || !connectServerForm || !serverTypeSelect || !stdioOptionsDiv || !sseOptionsDiv || !imageUpload || !imagePreviewContainer || !imagePreview || !removeImageBtn) {
+    if (!chatLog || !messageInput || !sendButton || !resetButton || !statusIndicator || !connectServerButton || !modal || !connectServerForm || !serverTypeSelect || !stdioOptionsDiv || !sseOptionsDiv || !imageUpload || !imagePreviewContainer || !imagePreview || !removeImageBtn) {
         console.error("Initialization failed: One or more required DOM elements not found.");
         // Display error to user if possible
         if (chatLog) {
@@ -45,7 +43,7 @@ document.addEventListener('DOMContentLoaded', () => {
     }
 
     function scrollToBottom() {
-       messageListWrapper.scrollTop = messageListWrapper.scrollHeight;
+        chatLog.scrollTop = chatLog.scrollHeight;
     }
 
     function displaySystemMessage(text, type = 'info') {
@@ -77,7 +75,7 @@ document.addEventListener('DOMContentLoaded', () => {
         }
 
         chatLog.appendChild(messageElement);
-        messageListWrapper.scrollTop = messageListWrapper.scrollHeight;
+        scrollToBottom();
     }
 
     function appendExpandableMessage(headerHtml, contentHtml, type) {
@@ -121,13 +119,13 @@ document.addEventListener('DOMContentLoaded', () => {
     }
 
     function isScrolledToBottom() {
-        const threshold = 5; 
-        return messageListWrapper.scrollHeight - messageListWrapper.scrollTop - messageListWrapper.clientHeight <= threshold;
+        const threshold = 5;
+        return chatLog.scrollHeight - chatLog.scrollTop - chatLog.clientHeight <= threshold;
     }
 
     function adjustTextareaHeight() {
-        userInput.style.height = 'auto'; // Reset height to calculate new height
-        userInput.style.height = `${userInput.scrollHeight}px`; // Set to scroll height
+        messageInput.style.height = 'auto';
+        messageInput.style.height = `${messageInput.scrollHeight}px`;
 
         // Adjust chat log padding based on input area height
         const inputAreaHeight = document.getElementById('input-area').offsetHeight;
@@ -137,7 +135,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
     // Modify sendMessage to include image data
     async function sendMessage() {
-        const messageText = userInput.value.trim();
+        const messageText = messageInput.value.trim();
 
         // Require text input even if image is present
         if (!messageText) {
@@ -147,15 +145,14 @@ document.addEventListener('DOMContentLoaded', () => {
         } 
 
         appendMessage('user', messageText, currentImageData ? currentImageData.base64 : null);
-        userInput.value = '';
-        userInput.style.height = 'auto'; // Reset height before sending
-        const messageToSend = messageText;
+        messageInput.value = '';
+        messageInput.style.height = 'auto'; // Reset height before sending
         const imageDataToSend = currentImageData; 
         removeImage(); // Clear image after grabbing data for sending
 
         // Send via WebSocket instead of fetch
         if (ws && ws.readyState === WebSocket.OPEN) {
-            const payload = { type: 'message', content: messageToSend, imageData: imageDataToSend };
+            const payload = { type: 'message', content: messageText, imageData: imageDataToSend };
             ws.send(JSON.stringify(payload));
         } else {
             displaySystemMessage('Cannot send message: Not connected to server.', 'error');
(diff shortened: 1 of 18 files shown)


shaunak99:
#38

src/ai/llm/services/vercel.ts
```diff
@@ -74,9 +75,13 @@ export class VercelLLMService implements ILLMService {
         }, {});
     }
 
-    async completeTask(userInput: string): Promise<string> {
-        // Add user message
-        this.messageManager.addUserMessage(userInput);
+    async completeTask(userInput: string, imageData?: { image: string | Uint8Array | Buffer | ArrayBuffer | URL, mimeType?: string }): Promise<string> {
```
rahulkarajgikar: this `{ image: string | Uint8Array | Buffer | ArrayBuffer | URL, mimeType?: string }` is repeated in multiple places, let's make it a type/interface `ImageData` that we can reuse

app/web/client/script.js
```diff
@@ -24,22 +25,39 @@ document.addEventListener('DOMContentLoaded', () => {
     const serverTypeSelect = document.getElementById('server-type');
     const stdioOptionsDiv = document.getElementById('stdio-options');
     const sseOptionsDiv = document.getElementById('sse-options');
+    const chatLog = document.getElementById('message-list');
```
rahulkarajgikar: `chatLog` is a duplicate of `messageList`

maybe we can keep `chatLog` - it's a more descriptive name

app/web/client/script.js
```diff
@@ -24,22 +25,39 @@ document.addEventListener('DOMContentLoaded', () => {
...
-             messageList.appendChild(errorElement);
+             chatLog.appendChild(errorElement);
         }
         return; // Stop further execution
     }
 
     // --- Helper Functions (Defined inside DOMContentLoaded) ---
 
+    // Function to update the status indicator
+    function updateStatus(text, className) {
```
rahulkarajgikar: what does status mean? this function is kinda vague

src/ai/llm/services/vercel.ts
```diff
@@ -166,37 +171,155 @@ export class VercelLLMService implements ILLMService {
                     `Step finished, step tool results: ${JSON.stringify(step.toolResults, null, 2)}`
                 );
 
-                // Process tool calls
+                if (step.text) {
+                    this.eventEmitter.emit('response', step.text);
+                }
```
rahulkarajgikar: why are we adding this here? `response` previously meant the final response, looks like this is changing the meaning of it

src/ai/llm/services/vercel.ts
```diff
@@ -166,37 +171,155 @@ export class VercelLLMService implements ILLMService {
...
+                        break;
+                }
+            }
+        }
 
-        // Add final assistant message
-        this.messageManager.addAssistantMessage(fullResponse);
+        const fullResponse = response.text; // Final text response remains the same
 
-        this.eventEmitter.emit('response', fullResponse);
```
rahulkarajgikar: why are we removing this - similar q to one of the earlier comments

src/ai/llm/services/vercel.ts
```diff
@@ -166,37 +171,155 @@ export class VercelLLMService implements ILLMService {
...
+                            this.messageManager.setSystemPrompt(message.content);
+                        }
+                        break;
+                    default:
+                        // Use 'role' variable which has the asserted type
+                        logger.warn(`[VercelLLMService] Unknown message role in response: ${role}`);
+                        break;
+                }
+            }
+        }
```
rahulkarajgikar: 1. this function is doing too much. we should move all of this response parsing to a separate function - `generateText` should just generate text and call another function to handle this parsing. single responsibility principle - in general if a function is doing too much or is more than 10-20 lines, it's good to break it down into smaller pieces, it improves readability and testability.
2. this code has a lot of nested if + switch statements making it hard to read, would help if we clean this up
3. i don't think we should allow updating system prompt from here so the case 'system' is probably unnecessary
4. let's avoid `any` types unless absolutely necessary - it's very hard to work with `any` types
5. we could consider adding a new `outputFormatter` interface into each LLMService that handles getting the response from the LLM and passing it to the MessageManager. It can be initialized in each LLMService.

This could then look like:

```
func generateText - calls generateText, then calls outputFormatter.handleResponse
```

```
func handleResponse - handles sending the messages into the messageManager. we can have separate output formatters for each LLMService, similar to how we currently have input formatters. this can call sub-functions:

func handleResponse:
   case 'user':
         handleUserResponse() - handles adding the user message into messagemanager
   case 'assistant'
         handleAssistantResponse() - handles adding the assistant message into messagemanager
.... and so on

```

and outputformatter/llmresponseformatter:

```
interface OutputFormatter:
        MessageManager messagemanager;
        func formatLLMResponse();
```

```
class VercelOutputFormatter implements OutputFormatter - for vercel llm service
class OpenAIOutputFormatter implements OutputFormatter - for openai llm service
class AnthropicOutputFormatter implements OutputFormatter - for anthropic llm service
```


this is more modular and would extend better to our other LLMServices

src/ai/llm/services/vercel.ts
```diff
@@ -74,9 +75,13 @@ export class VercelLLMService implements ILLMService {
...
-    async completeTask(userInput: string): Promise<string> {
-        // Add user message
-        this.messageManager.addUserMessage(userInput);
+    async completeTask(userInput: string, imageData?: { image: string | Uint8Array | Buffer | ArrayBuffer | URL, mimeType?: string }): Promise<string> {
+        // Add user message - use multimodal method if imageData is provided
+        if (imageData) {
+            this.messageManager.addUserMultimodalMessage(userInput, imageData.image, imageData.mimeType);
+        } else {
+            this.messageManager.addUserMessage(userInput);
+        }
```
rahulkarajgikar: we shouldn't have this if-else tree here. `LLMService` shouldn't handle deciding what to do with messages, `messagemanager` should handle this.

`addUserMessage` should handle all of this internally so that our llmService is clean and just calls `MessageManager`.

`addUserMultiModalMessage` can be an internal function within `MessageManager` that `addUserMessage` can call if needed

This way LLMServices only need to call `addUserMessage` and the `MessageManager` will handle everything

final code can look like this:

```
# inside LLMservice - unchanged from before, just now adding imageData [could be null or not]
this.messageManager.addUserMessage(userinput, imageData)
```

and 

```
func addUserMessage(userInput, imageData?) can handle internally figuring out what to do - we may not even need a separate 'addUserMultiModalMessage` if it's just the one if condition

     if imageData - then append as text, image array. 
     else append message as text (same as before)
```

src/ai/llm/messages/utils.ts
```diff
@@ -7,6 +7,12 @@ const DEFAULT_OVERHEAD_PER_MESSAGE = 4;
 /**
  * Counts the total tokens in an array of InternalMessages using a provided tokenizer.
  * Includes an estimated overhead per message.
+ * 
+ * NOTE: This function counts tokens on the raw InternalMessage history and has limitations:
+ * 1. It does not account for provider-specific formatting (uses raw content).
+ * 2. It ignores the token cost of images in multimodal messages (counts text only).
+ * 3. The overhead is a fixed approximation.
+ * For more accurate counting reflecting the final provider payload, use MessageManager.countTotalTokens().
```
rahulkarajgikar, line 15: this will cause problems if images take lots of tokens, we can't ignore tokens consumed by images - context window will fill up if there are a lot of images uploaded and our tokenizer won't count the tokens correctly so we wont compress the messages when we need to, and the llmservice could randomly fail due to too much context

for now we need at least some approximation for tokens consumed by images - maybe there is a way to do this based on the resolution/size of the image?

src/ai/llm/messages/utils.ts
```diff
@@ -23,20 +29,32 @@ export function countMessagesTokens(
...
+                    // For multimodal array content, count only text parts
+                    message.content.forEach(part => {
+                        if (part.type === 'text' && typeof part.text === 'string') {
+                            total += tokenizer.countTokens(part.text);
+                        }
+                        // Image tokens are ignored in this utility
+                    });
+                }
+                // else: Handle other potential content types if necessary in the future
             }
```
rahulkarajgikar: ok i'm realizing - `countMessagesTokens` in the utils file does the same thing as `countTokens` in the `messagemanager` so we have duplicate code

i'm thinking we can do this:

1. get rid of `countTotalTokens()` in utils and just keep one function countTokens() in the `MessageManager` `countTotalTokens()` util is currently used to count messages in the compression strategies before compression-but i think we can delete this check from the compression utils. [not in this PR diff btw].  tokens are already counted before we compress the messages. [this is already true]. this also helps with single responsibility principle formatter handles formatting inputs, tokenizer handles counting tokens from any message, compression util handles compression
3. With the new change, the countTokens() method in the llmservice is using the formatter to format and then has all the logic specific to vercelLLMService - offload this to service specific tokenizers - ties in with the previous point.

src/ai/llm/messages/manager.ts
```diff
@@ -226,42 +272,82 @@ export class MessageManager {
...
-                            total += this.tokenizer.countTokens(call.function.arguments);
-                        }
-                    }
+                // Count tokens for tool calls if present in the formatted message
+                // Vercel SDK uses 'toolCalls' or 'tool_calls' depending on context/version
+                const toolCalls = formattedMessage?.toolCalls || formattedMessage?.tool_calls;
+                if (toolCalls) {
+                     // Stringify to approximate token cost, as structure can vary
+                    const toolCallsString = JSON.stringify(toolCalls);
+                    totalTokens += this.tokenizer.countTokens(toolCallsString);
```
rahulkarajgikar: do we need this change in logic for counting tokens in tool calls? 

also we should stick to `toolCalls` or `tool_calls` and not both - whichever one is the one our version of vercel sdk actually uses. if vercel's name changes in a new version we can bump up our version as well and change this - no need to support both

src/ai/llm/messages/manager.ts
```diff
@@ -226,42 +272,82 @@ export class MessageManager {
...
-            if (this.systemPrompt) {
-                total += this.tokenizer.countTokens(this.systemPrompt);
-            }
-
-            // Count each message
-            for (const message of this.history) {
-                if (message.content) {
-                    total += this.tokenizer.countTokens(message.content);
+            // Get the full message history formatted for the target LLM provider
+            const formattedMessages = this.formatter.format(this.history, this.systemPrompt);
```
rahulkarajgikar: this token counting logic seems tied down to `VercelLLMService` -  `AnthropicService` or `OpenAIService` won't work accurately because the formatter works differently

formatting here does help with getting a more accurate representation of what really goes into the LLM so we should keep that.

but i think all  this logic of parsing the messages and passing it to the tokenizer below should live in the `tokenizer.countTokens()` method and the MessageManager should offload this work to it

then Messagemanager countTotalTokens() code to look very simple:

```
formattedMessages = this.formatter.format(this.history, this.systemPrompt);
for message in formattedMessages:
      this.tokenizer.countTokens(message)
```

and the `tokenizer.countTokens()` will anyway be llmservice specific, so it can know the expected format and figure out how to count the tokens

rahulkarajgikar (changes requested):
added many refactoring suggestions

connect github MCP and ask o4-mini/gpt-4.1 to give opinions/suggestions on the comments of this PR. it can validate if the comment is good and suggest fixes

src/ai/llm/services/vercel.ts
shaunak99: refer above

src/ai/llm/services/vercel.ts
shaunak99: The LLM may generate a text response along with a tool call in a single turn. The final text response is a special case where only text is generated without any tool calls. By moving it into onStepFinish we are able to capture and display any text content along with the tool call. The final response is also captured by default due to this.

app/web/client/script.js
shaunak99: Leftover code. Will remove

src/ai/llm/services/vercel.ts
shaunak99: working on this

shaunak99: Have updated this to be managed by the formatter

src/ai/llm/services/vercel.ts
rahulkarajgikar: Ah okay that makes sense, let's check if we are consuming this `response` event in the CLI or UI

We might have to make changes there as well (i guess mostly not though)

src/ai/llm/messages/formatters/vercel.ts
```diff
@@ -43,29 +43,101 @@ export class VercelMessageFormatter implements IMessageFormatter {
...
+
                         formatted.push({
                             role: 'assistant',
-                            content: msg.content,
-                            function_call: {
-                                name: msg.toolCalls[0].function.name,
-                                arguments: msg.toolCalls[0].function.arguments,
-                            },
+                            content: contentParts,
+                            // Remove deprecated function_call
```
rahulkarajgikar: is this actually deprecated? we didn't bump up version of any dependency so i think this shouldn't change

src/ai/llm/messages/formatters/vercel.ts
```diff
@@ -43,29 +43,101 @@ export class VercelMessageFormatter implements IMessageFormatter {
...
+                            imageDataBase64 = Buffer.from(imagePart.image).toString('base64');
+                        } else if (imagePart.image instanceof ArrayBuffer) {
+                           imageDataBase64 = Buffer.from(imagePart.image).toString('base64');
+                        } else if (imagePart.image instanceof URL) {
+                           imageDataBase64 = imagePart.image.toString(); // Pass URL string
+                        } else {
+                            // Fallback or throw error if type is unexpected
+                            console.warn('Unexpected image data type in Vercel formatter for tool result:', typeof imagePart.image);
+                            imageDataBase64 = '[Unsupported image data type]';
+                        }
```
rahulkarajgikar: we can move this to a separate function `getImageData(imagePart, base='base64'): string // default base64 for base` that handles this if else tree

src/ai/llm/messages/formatters/vercel.ts
```diff
@@ -43,29 +43,101 @@ export class VercelMessageFormatter implements IMessageFormatter {
...
+                                toolCallId: toolCall.id,
+                                toolName: toolCall.function.name,
+                                // Ensure args are parsed if they are a string, otherwise pass as is
+                                // The SDK expects args as an object
+                                args: typeof toolCall.function.arguments === 'string'
+                                        ? JSON.parse(toolCall.function.arguments)
+                                        : toolCall.function.arguments,
+                            });
+                        }
+
```
rahulkarajgikar: i'm thinking we can make this into a separate function `formatAssistantMessage` and similarly split it for `formatToolMessage` because these functions are a bit long. These 2 new functions can be internal to `vercel.ts`

will make it more testable and easier to read

```
                case 'user':
                case 'system':
                    formatted.push({
                        role: msg.role,
                        content: msg.content,
                    });
                    break;

                case 'assistant':
                         formatted.push({
                                role: 'assistant',
                                content:  formatAssistantMessage(msg)
                         })
                case 'tool':
                         formatted.push({
                                role: 'assistant',
                                content:  formatToolMessage(msg)
                         })
```

src/ai/llm/messages/formatters/vercel.ts
rahulkarajgikar: so this `getImageData` could be called in the new `formatToolMessage`

src/ai/llm/messages/formatters/vercel.ts
```diff
@@ -84,4 +156,67 @@ export class VercelMessageFormatter implements IMessageFormatter {
     getSystemPrompt(): null {
         return null;
     }
+
+    /**
+     * Parses raw Vercel SDK response into internal message objects.
+     */
+    parseResponse?(response: any): InternalMessage[] {
```
rahulkarajgikar: let's not use `any` type, find out the type is using for the response. it will show if you hover over the actual function that first generated this response. probably `generateText`. hard to understand what this function does without types

src/ai/llm/messages/formatters/vercel.ts
```diff
@@ -84,4 +156,67 @@ export class VercelMessageFormatter implements IMessageFormatter {
     getSystemPrompt(): null {
         return null;
     }
+
+    /**
+     * Parses raw Vercel SDK response into internal message objects.
+     */
```
rahulkarajgikar, line 76: let's add a TODO comment to break this down into smaller functions, it's hard to understand what's going on and there's probably scope for making this cleaner but we can do it later

src/ai/llm/services/vercel.ts
```diff
@@ -166,38 +170,38 @@ export class VercelLLMService implements ILLMService {
...
 
-        const fullResponse = response.text;
-
-        // Add final assistant message
-        this.messageManager.addAssistantMessage(fullResponse);
-
-        this.eventEmitter.emit('response', fullResponse);
-        return fullResponse;
+        // Parse and append each new InternalMessage from the formatter
+        if (typeof this.formatter.parseResponse === 'function') {
```
rahulkarajgikar: check this

src/ai/llm/messages/formatters/types.ts
```diff
@@ -17,6 +17,11 @@ export interface IMessageFormatter {
      */
     format(history: Readonly<InternalMessage[]>, systemPrompt?: string | null): any[];
 
+    /**
+     * Parses raw LLM response into an array of InternalMessage objects.
+     */
+    parseResponse?(response: any): InternalMessage[];
```
rahulkarajgikar: should be mandatory

src/ai/llm/messages/manager.ts
```diff
@@ -190,13 +202,35 @@ export class MessageManager {
...
+
+        let content: InternalMessage['content'];
+
+        // --- Check for Image Data ---
+        // Adjust this condition based on how your MCP server actually returns images
+        if (
+            result &&
+            typeof result === 'object' &&
+            typeof result.contentType === 'string' &&
+            result.contentType.startsWith('image/') &&
```
rahulkarajgikar: is this hardcoded for a specific mcp server?

src/ai/llm/services/vercel.ts
shaunak99: Fixed, removed the if condition

src/ai/llm/messages/formatters/vercel.ts
shaunak99: Good catch, added it back!

src/ai/llm/messages/manager.ts
shaunak99: buggy implementation, have fixed it

src/ai/llm/messages/manager.ts
```diff
@@ -219,49 +245,19 @@ export class MessageManager {
     }
 
     /**
-     * Counts total tokens in current conversation including system prompt.
-     * Requires a tokenizer to be configured.
-     * @returns Total token count or null if no tokenizer is available.
+     * Counts total tokens in current InternalMessage history using the shared util.
+     * @returns Total token count or null if tokenizer is not set or counting fails.
      */
     countTotalTokens(): number | null {
```
rahulkarajgikar: delete this function

app/web/client/script.js
```diff
@@ -4,14 +4,14 @@ import { marked } from 'https://cdn.jsdelivr.net/npm/marked/lib/marked.esm.js';
...
 
 // Configure marked with sanitization (IMPORTANT!)
 marked.use({ sanitize: true });
 
 // --- Wait for DOMContentLoaded before initializing UI and listeners ---
 document.addEventListener('DOMContentLoaded', () => {
     // --- Get DOM Elements --- 
-    const messageList = document.getElementById('message-list');
     const messageListWrapper = document.getElementById('message-list-wrapper');
     const messageInput = document.getElementById('message-input');
```
rahulkarajgikar, line 16: also duplicate

src/ai/llm/messages/formatters/openai.ts
```diff
@@ -87,4 +87,12 @@ export class OpenAIMessageFormatter implements IMessageFormatter {
     getSystemPrompt(): null {
         return null;
     }
+
+    /**
+     * Parses OpenAI API response into internal message objects.
+     * Currently a no-op stub; implement as needed.
+     */
+    parseResponse(response: any): InternalMessage[] {
```
rahulkarajgikar, line 95: needs to be implemented

src/ai/llm/messages/formatters/anthropic.ts
```diff
@@ -170,4 +170,12 @@ export class AnthropicMessageFormatter implements IMessageFormatter {
         // Anthropic uses system prompt as a separate parameter, no need for any formatting
         return systemPrompt;
     }
+
+    /**
+     * Parses Anthropic API response into internal message objects.
+     * Currently a no-op stub; implement actual parsing logic as needed.
+     */
+    parseResponse(response: any): InternalMessage[] {
```
rahulkarajgikar, line 178: needs to be implemented

src/client/mcp-client.ts
```diff
@@ -175,7 +175,7 @@ export class MCPClient implements ToolProvider {
             await this.client.connect(this.transport);
             // If connection is successful, we know the server was spawned
             this.serverSpawned = true;
-            logger.info(`✅ SSE SERVER ${url} SPAWNED`);
```
rahulkarajgikar, line 178: add back

shaunak99: added serverName instead

src/ai/llm/messages/utils.ts
rahulkarajgikar: going with approximation for now

src/ai/llm/messages/utils.ts
rahulkarajgikar: putting this in utils for now but might move back

rahulkarajgikar (approved):

PR #65, opened 2025-04-22
Tool approval refactors

rahulkarajgikar:
Minor changes as follow up to #43 

1. Renaming some files and variables - renamed `FileSettingsProvider` to InMemorySettingsProvider to be more accurate. renamed `UserConfirmationProvider` to `ToolConfirmationProvider` to be more accurate to its purpose
2. Adding a `user-info.ts` utility to fetch user ID - defaults to a constant right now but can be improved later
3. Replaced hardcoded set of `allowedTools` in `UserConfirmationProvider` with an `AllowedToolsProvider` interface - this allows us to extend to files/db/etc. `AllowedToolsProvider` supports multi-tenancy but defaults to single tenancy. This enables a route to multi-tenancy.
4. Removed `userId` from `userSettings` interface - this interface should hold the settings for a specific user-id but shouldn't store the actual user id (more representative of actual files/dbs)

Todo:
1. expose these settings all the way to the top to make them configurable (CLI/UI/saiki-config)
2. Add more sub-settings - like ToolSettings within the UserSettings, AppSettings for overall app level settings. Goal is to have everything configurable
3. Start using the callbacks/event emitter properly


src/client/manager.ts
```diff
@@ -196,14 +190,6 @@ export class ClientManager {
...
     }
 
-    /**
-     * Get all allowed tools
-     * @returns Set of tool names that are allowed without confirmation
-     */
-    getAllowedTools(): Set<string> {
-        return this.confirmationProvider.allowedTools;
-    }
-
```
rahulkarajgikar, line 206: removing this because this just routes to a dependency

this will only be needed in the CLI/UI.

We will move all initialization to the root of the project in the CLI/UI/wherever and expose APIs/functions as needed on top - best practices for dependency injection. improves testability, configurability and readability

Issue #67, opened 2025-04-22
Refactor codebase to allow complete configurability


rahulkarajgikar:
Right now we expose some configurations for MCP and LLM configs in `saiki.yml`. 

Goal: For any new feature we add, we HAVE to be able to configure it through this config file. This gives users maximum configurability without having to change much code.

Problem: The codebase isn't well set up at the moment because we aren't utilizing dependency injection properly.

Solution: 
- The best practice for doing that is to initialize all important/shared services at the beginning of the execution of the program. This would be our `initializeServices` function. 
- This also means we should ALWAYS accept dependencies through constructor arguments and NEVER initialize dependencies inside constructors.
- This involves moving things like `EventEmitter`, `MessageManager`, `ClientManager`,`LLMService` all to be initialized in one starting point `initializeServices`. Some of these services are currently initialized inside other services, making them less re-usable in new places.
- For example: Suppose tomorrow we need a new class that requires the event bus, the messsage manager, and the tokenizer. Today, there is no straightforward way to do this because we are initializing some of these services inside other services. If all services were created in `initializeServices` this would be trivial

rahulkarajgikar:
done in #70

PR #70, opened 2025-04-23
Move service initializations into service-initializer.ts

rahulkarajgikar:
This will help with #67 . This follows best practices of dependency injection - inject all dependencies from the top and helps us achieve the following goals:

1. Easier configurablity - now that services are being initialized at the top, it's easy to wire options/configs from `saiki.yml` to decide how to initialize them and give maximum control to users.
2. More testable and modular - dependency injection allows easily substituting mock classes instead of real ones wherever needed.
3. Direct access to services for web UI/CLI interfaces - having services at the top also makes it much easier to add features for our web UI and CLI to make it more useful

### High level Changes:
1. Have all services and dependencies initialized at the top level - `service-initializer.ts`
4. Allow us to configure most features via config file `saiki.yml`
5. Have a way to enable internal overrides through code - this is useful for testing and any custom use-cases.
6. Wherever we needed a way to control what specific implementation of an interface is created, we have added a `factory.ts` file in the specific folder with the required arguments.

On 1 - we pass the `AgentConfig` into `service-initializer.ts`. We can update the logic of this file to read the appropriate sub-configs from `AgentConfig` and use that to initialize services in a custom manner

On 2 - created `InitializeServiceOptions` - which goes into `service-initializer.ts`. This allows us to have code-level overrides for top level services like LLMService, MessageManager, etc. 

Priority: code override > config file > internal defaults

### Tenets
1. Allow all features to be configurable via either config file or code-level override
2. Allow low-level classes/services (Formatters/tokenizers/etc.) to be configurable via config file
3. Only keep high level classes/services (LLMService, MessageManager, etc.) to be configurable via code-override - this is to avoid bloating `InitializeServiceOptions`

This enables us to use config file to support most features and low level customization, but for advanced code level requirements (like custom testing and mocking), there is still flexibility to inject top level services like `LLMService`, `ClientManager`, `MessageManager`, etc along with their low level dependencies.

### Notes:

1. As of now, we haven't updated our `AgentConfig` yet to allow these things to be customizable via config file - this will come later.
2. We also haven't fully exposed low level classes like `Formatters`, `Tokenizers`, `ToolConfirmationProvider` to be driven via config. This hasn't been done because these are mostly internal implementations. If we do need to make these also configurable, it will be much easier now that initialization is closer to the top. (just add a new section in `AgentConfig`, and update the corresponding `factory` or `initializeServices` logic)

### DIFF
--- a/app/index.ts
+++ b/app/index.ts
@@ -108,7 +108,7 @@ async function startAgent() {
         logger.info('===============================================\n');
 
         // Use the shared initializer
-        const { clientManager, llmService, agentEventBus } = await initializeServices(config, connectionMode);
+        const { clientManager, llmService, agentEventBus } = await initializeServices(config, { connectionMode });
 
         // Start based on mode
         if (runMode === 'cli') {
--- a/app/index.ts
+++ b/app/index.ts
@@ -108,7 +108,7 @@ async function startAgent() {
         logger.info('===============================================\n');
 
         // Use the shared initializer
-        const { clientManager, llmService, agentEventBus } = await initializeServices(config, { connectionMode });
+        const { clientManager, llmService, agentEventBus } = await initializeServices(config, { connectionMode, runMode });
 
         // Start based on mode
         if (runMode === 'cli') {
--- a/app/index.ts
+++ b/app/index.ts
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 import { existsSync } from 'fs';
 import { Command } from 'commander';
-import path from 'path';
 import dotenv from 'dotenv';
 import { logger } from '../src/utils/logger.js';
 import { loadConfigFile } from '../src/config/loader.js';
--- a/src/ai/llm/messages/manager.ts
+++ b/src/ai/llm/messages/manager.ts
@@ -356,4 +356,24 @@ export class MessageManager {
             }
         }
     }
+
+    /**
+     * Parses a raw LLM response, converts it into internal messages and adds them to the history.
+     *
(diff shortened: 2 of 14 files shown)


src/ai/llm/messages/manager.ts
```diff
@@ -54,30 +54,65 @@ export class MessageManager {
...
         this.maxTokens = maxTokens;
         this.tokenizer = tokenizer;
         this.compressionStrategies = compressionStrategies;
     }
 
+    /**
+     * Returns the current token count of the conversation history.
+     * @returns The number of tokens in the current history
+     */
+    getTokenCount(): number {
```
rahulkarajgikar, line 86: added this back for llmservice to have a way to get current token count without needing the tokenizer/formatter

app/index.ts
```diff
@@ -108,7 +107,7 @@ async function startAgent() {
         logger.info('===============================================\n');
 
         // Use the shared initializer
-        const { clientManager, llmService, agentEventBus } = await initializeServices(config, connectionMode);
+        const { clientManager, llmService, agentEventBus } = await initializeServices(config, { connectionMode, runMode });
```
shaunak99, line 110: would it be better to make it into a single options parameter instead of (config, options)?

shaunak99 (approved):

app/index.ts
rahulkarajgikar: for now let's keep them separate - config and overrides would be independent

in the future we could combine them before, if the config gets too large

PR #76, opened 2025-04-28
Use saiki in headless mode

shaunak99:
This PR enable users to run commands like `saiki "can you tell me what are LLMs?"`


rahulkarajgikar:
@CodiumAI-Agent /review

app/cli/cli.ts
```diff
@@ -150,3 +150,44 @@ export async function runAiCli(
...
+    agentEventBus.on('llmservice:error', cliSubscriber.onError.bind(cliSubscriber));
+    agentEventBus.on('llmservice:conversationReset', cliSubscriber.onConversationReset.bind(cliSubscriber));
+
+    // Load tools and confirm initialization
+    logger.info('Loading available tools...');
+    const tools = await clientManager.getAllTools();
+    logger.info(
+        `Loaded ${Object.keys(tools).length} tools from ${clientManager.getClients().size} MCP servers\n`
+    );
+    logger.info('AI Agent initialized successfully!', null, 'green');
```
rahulkarajgikar: this code [everything before  line 192] is the same setup as the one we have in `runAiCLi`

we can make a new function `initCli` that does this logging + subscription + etc. setup before we call `llmservice.completeTask`. we can use this function in both `runAiCli` and `runHeadlessCli` to reduce code duplication

rahulkarajgikar (changes requested):

rahulkarajgikar (approved):

PR #78, opened 2025-04-29
0.2.1

rahulkarajgikar:
added version 0.2.1 with `npm version patch` - testing automated publishing


rahulkarajgikar:
update: don't do this via PR because it doesn't merge git tags, directly use git push to push to upstream

PR #82, opened 2025-04-29
Added support for MCP resources prompts

shaunak99:
MCP client has been updated to support resources and prompts

Added a max_iterations parameter to control total iterations of tool calling loops

Added o4-mini support to model options


app/index.ts
```diff
@@ -52,7 +52,7 @@ program
     .option('-c, --config-file <path>', 'Path to config file', DEFAULT_CONFIG_PATH)
     .option('-s, --strict', 'Require all server connections to succeed')
     .option('--no-verbose', 'Disable verbose output')
-    .option('--mode <mode>', 'Run mode: cli or web', 'cli')
+    .option('--mode <mode>', 'Run mode: cli, web, or gameboy', 'cli')
```
rahulkarajgikar: remove gameboy for now

src/ai/llm/services/factory.ts
```diff
@@ -55,6 +55,8 @@ function _createInBuiltLLMService(
 ): ILLMService {
     // Extract and validate API key
     const apiKey = extractApiKey(config);
+    // Determine max tool iteration steps (default to 10)
+    const maxIter = config.maxIterations ?? 10;
```
rahulkarajgikar: keep it `maxIterations`. also let's remove the fallback to 10, we should rely on the defaults defined in the respective llmservice constructors - this allows different defaults for each llm service

src/ai/llm/services/openai.ts
```diff
@@ -58,11 +61,11 @@ export class OpenAIService implements ILLMService {
         this.eventEmitter.emit('llmservice:thinking');
 
         // Maximum number of tool use iterations
-        const MAX_ITERATIONS = 10;
+        const maxIterations = this.maxIterations;
```
rahulkarajgikar: we can just directly use `this.maxIterations` instead of making a new constant

src/ai/llm/services/factory.ts
```diff
@@ -102,7 +106,9 @@ function _createVercelLLMService(
     messageManager: MessageManager
 ): VercelLLMService {
     const model: LanguageModelV1 = _createVercelModel(config.provider, config.model);
-    return new VercelLLMService(clientManager, model, agentEventBus, messageManager);
+    // Allow overriding of maxIterations via config.llm.maxIterations
+    const maxIter = config.maxIterations ?? 50;
```
rahulkarajgikar: same as above

src/ai/llm/tokenizer/utils.ts
```diff
@@ -10,6 +10,7 @@ const MODEL_TOKEN_LIMITS: Record<string, Record<string, number>> = {
         'gpt-4.1': 1000000,
         'gpt-4o-mini': 128000,
         'gpt-4o': 128000,
+        'o4-mini': 128000,
```
rahulkarajgikar: i think this might be 200k: https://platform.openai.com/docs/models/o4-mini

src/client/types.ts
```diff
@@ -34,11 +34,11 @@ export interface IMCPClientWrapper extends ToolProvider {
     // Connection Management
     connect(config: McpServerConfig, serverName: string): Promise<IMCPClient>;
 
-    // TODO: implement Prompt Management
-    // listPrompts(): Promise<string[]>;
-    // getPrompt(name: string, args?: any): Promise<string>;
+    // Prompt Management
+    listPrompts(): Promise<string[]>;
+    getPrompt(name: string, args?: any): Promise<any>; // Return type might need refinement based on SDK
```
rahulkarajgikar: return type for prompts/get looks to be defined with this structure:

https://modelcontextprotocol.io/docs/concepts/prompts: 
```
// Response
{
  description: "Analyze Python code for potential improvements",
  messages: [
    {
      role: "user",
      content: {
        type: "text",
        text: "Please analyze the following Python code for potential improvements:\n\n```python\ndef calculate_sum(numbers):\n    total = 0\n    for num in numbers:\n        total = total + num\n    return total\n\nresult = calculate_sum([1, 2, 3, 4, 5])\nprint(result)\n```"
      }
    }
  ]
}
```

mcp sdk probably has a type for this which we should import

src/client/types.ts
```diff
@@ -34,11 +34,11 @@ export interface IMCPClientWrapper extends ToolProvider {
...
 
     // Resource Management
-    // listResources(): Promise<string[]>;
-    // readResource(url: string): Promise<string>;
+    listResources(): Promise<string[]>;
+    readResource(uri: string): Promise<any>; // Return type might need refinement based on SDK
```
rahulkarajgikar: any type is probably fine here since it's hard to say what resources would return when we read

src/ai/llm/services/vercel.ts
```diff
@@ -77,8 +80,7 @@ export class VercelLLMService implements ILLMService {
             `[VercelLLMService] Formatted tools: ${JSON.stringify(formattedTools, null, 2)}`
         );
 
-        // Maximum number of tool use iterations
-        const MAX_ITERATIONS = 50;
+        // Use configured maximum number of tool iterations
```
rahulkarajgikar: don't need this comment anymore

src/client/types.ts
```diff
@@ -34,11 +34,11 @@ export interface IMCPClientWrapper extends ToolProvider {
     // Connection Management
```
rahulkarajgikar, line 18: we should rename this to `IMCPClient` and delete the other unused IMCPClient and VercelMCPClient on top, no need to use the suffix `Wrapper` anymore

we also need to update `MCPClient` class to implement this interface and not the `ToolProvider` interface, right now this interface is unused

we should also move the `disconnect` method from `ToolProvider` interface into `IMCPClient`

src/client/manager.ts
```diff
@@ -92,10 +141,96 @@ export class ClientManager {
...
+        if (!client) {
+            logger.warn(`Prompt '${name}' not found in cache. Rebuilding cache and retrying.`);
+            await this.rebuildAllCaches();
+            const refreshedClient = this.getPromptClient(name);
+            if (!refreshedClient) {
+                throw new Error(`No client found for prompt: ${name} even after cache rebuild.`);
+            }
+            if (typeof refreshedClient.getPrompt !== 'function') {
+                 throw new Error(`Client found for prompt '${name}' does not support getPrompt.`);
+            }
```
rahulkarajgikar: not understanding this and similar checks - won't client.getPrompt always exist because it's defined in mcp-client.ts as a wrapper function?

this logic to see if the server actually supports the feature should be handled in the corresponding `getPrompt` function in mcp-client.ts

if the feature/capability isn't implemented in the server, we should throw some custom exception like `FeatureNotSupportedException` or something like that. MCP SDK probably has a way to vet this

or maybe in `mcp-client.ts` we might need some function that checks what capabilities the server actually supports

src/client/manager.ts
```diff
@@ -1,15 +1,17 @@
 import { MCPClient } from './mcp-client.js';
 import { ServerConfigs, McpServerConfig } from '../config/types.js';
 import { logger } from '../utils/logger.js';
-import { ToolProvider } from './types.js';
+import { IMCPClientWrapper } from './types.js';
 import { ToolConfirmationProvider } from './tool-confirmation/types.js';
 import { CLIConfirmationProvider } from './tool-confirmation/cli-confirmation-provider.js';
 import { ToolSet } from '../ai/types.js';
 
 export class ClientManager {
```
rahulkarajgikar: we can rename this class to MCPClientManager also

src/client/manager.ts
```diff
@@ -18,70 +20,117 @@ export class ClientManager {
...
+        logger.debug('Finished rebuilding all caches.');
+    }
+
+    /**
+     * Get all available tools from all connected clients, updating the cache.
+     * @returns Promise resolving to a ToolSet mapping tool names to Tool definitions
+     */
+    async getAllTools(): Promise<ToolSet> {
+        // Refresh all caches before retrieving tools
+        await this.rebuildAllCaches();
```
rahulkarajgikar: do we need to rebuild all caches every time - this could be expensive

src/client/manager.ts
```diff
@@ -92,10 +141,96 @@ export class ClientManager {
         if (!approved) {
             throw new Error(`Execution of tool '${toolName}' was denied`);
         }
-
+            return await refreshedClient.callTool(toolName, args);
+        }
```
rahulkarajgikar: indentation of this function is not right so the logic is changing - we are now bypassing the tool confirmation provider if the client was already there

the if(!client) block needs to end at line 133 and not line 145

correct code skeleton

```
async executeTool(toolName: string, args: any): Promise<any> {
        const client = this.getToolClient(toolName);
        if (!client) {
            logger.warn(`Tool '${toolName}' not found in cache. Rebuilding cache and retrying.`);
            await this.rebuildAllCaches();
            client = this.getToolClient(toolName);
            if (!client) {
                throw new Error(`No client found for tool: ${toolName} even after cache rebuild.`);
            }
        } // block ends here now

        // Request confirmation before executing
        const approved = await this.confirmationProvider.requestConfirmation({
            toolName,
            args,
        });

        if (!approved) {
            throw new Error(`Execution of tool '${toolName}' was denied`);
        }
        return await client.callTool(toolName, args);
}
```

rahulkarajgikar (changes requested):

src/client/manager.ts
```diff
@@ -18,70 +20,117 @@ export class ClientManager {
...
+            const tools = await client.getTools();
+            for (const toolName in tools) {
+                this.toolToClientMap.set(toolName, client);
+            }
 
-                // Map each tool to its provider client
-                for (const toolName in toolList) {
-                    this.toolToClientMap.set(toolName, client);
-                }
+            if (typeof client.listPrompts === 'function') {
```
rahulkarajgikar: this is the right idea but i don't think this is the right check - `client.listPrompts` maps to the listPrompts function defined in mcp-client.ts which will always be a functoin - so this will always be true

then if the client actually doesn't support prompts, we'll get back an exception we need to gracefully handle (some clients may not support prompts)

src/client/manager.ts
rahulkarajgikar: it looks like the MCP SDK doesn't return any custom exceptions if the capability isn't supproted

```
assertCapabilityForMethod(method) {
        var _a, _b, _c, _d, _e;
        switch (method) {
            case "logging/setLevel":
                if (!((_a = this._serverCapabilities) === null || _a === void 0 ? void 0 : _a.logging)) {
                    throw new Error(`Server does not support logging (required for ${method})`);
                }
                break;
            case "prompts/get":
            case "prompts/list":
                if (!((_b = this._serverCapabilities) === null || _b === void 0 ? void 0 : _b.prompts)) {
                    throw new Error(`Server does not support prompts (required for ${method})`);
                }
                break;
            case "resources/list":
            case "resources/templates/list":
            case "resources/read":
            case "resources/subscribe":
            case "resources/unsubscribe":
                if (!((_c = this._serverCapabilities) === null || _c === void 0 ? void 0 : _c.resources)) {
                    throw new Error(`Server does not support resources (required for ${method})`);
                }
                if (method === "resources/subscribe" &&
                    !this._serverCapabilities.resources.subscribe) {
                    throw new Error(`Server does not support resource subscriptions (required for ${method})`);
                }
                break;
            case "tools/call":
            case "tools/list":
                if (!((_d = this._serverCapabilities) === null || _d === void 0 ? void 0 : _d.tools)) {
                    throw new Error(`Server does not support tools (required for ${method})`);
                }
                break;
            case "completion/complete":
                if (!((_e = this._serverCapabilities) === null || _e === void 0 ? void 0 : _e.completions)) {
                    throw new Error(`Server does not support completions (required for ${method})`);
                }
                break;
            case "initialize":
                // No specific capability required for initialize
                break;
            case "ping":
                // No specific capability required for ping
                break;
        }
    }
```

they just throw a generic `Error` with a message. there is a common string `does not support` which we can check in the exception message which we catch to know if the reason for failure was becuase of unsupported capability - something like this:

```
try {
    await client.listPrompts();
} catch (error) {
    if (
        error instanceof Error &&
        error.message &&
        error.message.includes("Server does not support prompts")
    ) {
        // Handle unsupported capability
    } else {
        // Handle other errors
    }
}
```

this is not ideal though - i cut an issue to MCP repo to improve this: https://github.com/modelcontextprotocol/typescript-sdk/issues/430

rahulkarajgikar: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/index.ts - looks like they have a getCapabilities() function, that could also be a way to check the capabilities of a server

src/client/types.ts
shaunak99: Set return type as GetPromptResult from the SDK

src/client/types.ts
shaunak99: Set return type as ReadResourceResult from the SDK

src/client/manager.ts
shaunak99: definitely expensive, have removed cacheing implementation overall, can revisit it in the future. Leaving the cache-related functions in place, let me know if I should remove those as well.

src/client/manager.ts
shaunak99: remove cache, review updated code

src/client/manager.ts
shaunak99: Updated to try-catch

src/client/manager.ts
shaunak99: I have added a simple error to be thrown for now in the mcp-client, will need to update mcp sdk version of the SDK for getCapabilities()

src/ai/llm/services/anthropic.ts
```diff
@@ -58,12 +61,12 @@ export class AnthropicService implements ILLMService {
         this.eventEmitter.emit('llmservice:thinking');
 
         // Maximum number of tool use iterations
-        const MAX_ITERATIONS = 10;
+        const maxIterations = this.maxIterations;
```
rahulkarajgikar: use `this.maxIterations` directly, we don't need a constant that just maps to another constant

src/ai/llm/services/factory.ts
```diff
@@ -49,12 +49,13 @@ function extractApiKey(config: LLMConfig): string {
...
 function _createInBuiltLLMService(
     config: LLMConfig,
-    clientManager: ClientManager,
+    clientManager: MCPClientManager,
     agentEventBus: EventEmitter,
     messageManager: MessageManager
 ): ILLMService {
     // Extract and validate API key
     const apiKey = extractApiKey(config);
+    const maxIter = config.maxIterations
```
rahulkarajgikar: use `config.maxIterations` directly

src/ai/llm/services/factory.ts
```diff
@@ -97,12 +100,13 @@ function _createVercelModel(provider: string, model: string): LanguageModelV1 {
...
 function _createVercelLLMService(
     config: LLMConfig,
-    clientManager: ClientManager,
+    clientManager: MCPClientManager,
     agentEventBus: EventEmitter,
     messageManager: MessageManager
 ): VercelLLMService {
     const model: LanguageModelV1 = _createVercelModel(config.provider, config.model);
-    return new VercelLLMService(clientManager, model, agentEventBus, messageManager);
+    const maxIter = config.maxIterations
```
rahulkarajgikar: use `config.maxIterations` directly

src/client/manager.ts
```diff
@@ -82,20 +135,74 @@ export class ClientManager {
...
+    getPromptClient(promptName: string): IMCPClient | undefined {
+        return this.promptToClientMap.get(promptName);
+    }
+
+    /**
+     * Get a specific prompt definition by name.
+     * @param name Name of the prompt.
+     * @param args Arguments for the prompt (optional).
+     * @returns Promise resolving to the prompt definition.
+     */
```
rahulkarajgikar, line 155: change `any` return type, check below comments

src/client/manager.ts
```diff
@@ -18,61 +20,112 @@ export class ClientManager {
...
+                this.resourceToClientMap.set(resourceUri, client);
+            });
+            logger.debug(`Cached resources for client: ${clientName}`);
+        } catch (error) {
+            logger.debug(`Skipping resources for client ${clientName}: ${error}`);
+        }
+    }
+
+    //Not used at the moment
+    private async rebuildAllCaches(): Promise<void> {
```
rahulkarajgikar: seems like this isn't being used anywhere, we can delete it and add it back if/when we need it

src/client/manager.ts
```diff
@@ -137,15 +244,6 @@ export class ClientManager {
...
         }
-
-        if (successfulConnections.length >= requiredSuccessfulConnections) {
-            // Only clear errors related to the servers we attempted to connect to in this call
-            Object.keys(serverConfigs).forEach((name) => {
-                if (successfulConnections.includes(name)) {
-                    delete this.connectionErrors[name];
-                }
-            });
-        }
```
rahulkarajgikar, line 148: don't think we need to delete this

src/client/mcp-client.ts
```diff
@@ -2,12 +2,10 @@ import { Client } from '@modelcontextprotocol/sdk/client/index.js';
...
-import path from 'path';
-import { fileURLToPath } from 'url';
 import { logger } from '../utils/logger.js';
 import { McpServerConfig, StdioServerConfig, SSEServerConfig } from '../config/types.js';
 import { ToolSet } from '../ai/types.js';
-import { ToolProvider } from './types.js';
+import { IMCPClient } from './types.js';
 import { resolvePackagePath } from '../utils/path.js';
 
 const ToolsListSchema = z.object({
```
rahulkarajgikar: can delete this as well since we're not using it anymore

src/client/mcp-client.ts
```diff
@@ -261,22 +259,105 @@ export class MCPClient implements ToolProvider {
...
+            const listToolResult = await this.client.listTools({});
+            logger.debug(`listTools result: ${JSON.stringify(listToolResult, null, 2)}`);
+
+            // Assume listToolResult.tools exists and iterate
+            if (listToolResult && listToolResult.tools) {
+                listToolResult.tools.forEach((tool: any) => {
+                    tools[tool.name] = {
+                        description: tool.description || 'No description provided',
+                        // Map inputSchema from SDK to parameters in local Tool type
+                        parameters: tool.inputSchema || { type: 'object', properties: {} },
```
rahulkarajgikar: not sure if we should have defaults here for `description` and `parameters`, might lead to weird side effects that will be hard to debug. if the tool description or input schema isn't there it's probably better to have a hard failure rather than have tools act strangely.

src/client/types.ts
```diff
@@ -1,44 +1,29 @@
 import { McpServerConfig } from '../config/types.js';
 import { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import { ToolSet } from '../ai/types.js';
-
-// Defining types as any because we can't import the types from the SDK
-export type VercelMCPClient = any;
-
-export type IMCPClient = Client | VercelMCPClient;
+import { GetPromptResult, ReadResourceResult } from '@modelcontextprotocol/sdk/types.js';
```
rahulkarajgikar, line 4: nice! now that we have these types, let's also use them in `mcp-client.ts` instead of `any` for getPrompt and readResource

src/client/manager.ts
```diff
@@ -166,11 +264,13 @@ export class ClientManager {
             logger.info(`Attempting to connect to new server '${name}'...`);
             await client.connect(config, name);
             this.registerClient(name, client);
-            logger.info(`Successfully connected to new server '${name}'`);
+            await this.updateClientCache(name, client);
```
rahulkarajgikar, line 255: nice, this is a better place to update the cache

src/client/mcp-client.ts
```diff
@@ -261,22 +259,105 @@ export class MCPClient implements ToolProvider {
...
+    /**
+     * Get a specific prompt definition
+     * @param name Name of the prompt
+     * @param args Arguments for the prompt (optional)
+     * @returns Prompt definition (structure depends on SDK)
+     */
+    async getPrompt(name: string, args?: any): Promise<any> {
+        this.ensureConnected();
+        try {
+            logger.debug(`Getting prompt '${name}' with args: ${JSON.stringify(args, null, 2)}`);
```
rahulkarajgikar, line 308: useful debug logs - good addition

src/client/mcp-client.ts
```diff
@@ -261,22 +259,105 @@ export class MCPClient implements ToolProvider {
...
+     * @returns Array of available resource URIs
+     */
+    async listResources(): Promise<string[]> {
+        this.ensureConnected();
+        try {
+            const response = await this.client.listResources();
+            logger.debug(`listResources response: ${JSON.stringify(response, null, 2)}`);
+            return response.resources.map((r: any) => r.uri);
+        } catch (error) {
+            logger.debug('Failed to list resources from MCP server (optional feature), skipping:', error);
```
rahulkarajgikar: let's use `logger.error` and not `logger.debug` for any catch blocks

src/utils/service-initializer.ts
```diff
@@ -66,7 +66,7 @@ export type AgentServices = {
 export type InitializeServicesOptions = {
     runMode?: 'cli' | 'web'; // Context/mode override
     connectionMode?: 'strict' | 'lenient'; // Connection mode override
-    clientManager?: ClientManager; // Inject a custom or mock ClientManager
+    clientManager?: MCPClientManager; // Inject a custom or mock ClientManager
```
rahulkarajgikar: nit: update comment to say `MCPClientManager`

src/client/manager.ts
```diff
@@ -18,61 +20,112 @@ export class ClientManager {
     }
 
     /**
-     * Register a client that provides tools
+     * Register a client that provides tools (and potentially more)
      * @param name Unique name for the client
-     * @param client The tool provider client
+     * @param client The client instance, expected to be IMCPClientWrapper
```
rahulkarajgikar: nit: `IMCPClient`, not `IMCPClientWrapper`

rahulkarajgikar (changes requested):
minor comments, mostly LGTM

src/ai/llm/services/anthropic.ts
shaunak99: my bad

src/client/manager.ts
shaunak99: Reasoninig by cursor:

> That block was effectively a no-op once we moved all error clearing into registerClient/connectClient. Every time a client actually connects successfully, we already do
`delete this.connectionErrors[name];`
inside registerClient, and on failure we populate this.connectionErrors[name]. Keeping the extra “if successful >= required, loop through and delete errors” was just duplicating that same cleanup. So we removed it to DRY up the code.

rahulkarajgikar: ah okay

rahulkarajgikar (approved):

PR #90, opened 2025-05-05
Enhanced backend with a dedicated api/server layer

shaunak99:
Added a new api layer for consolidating all endpoints to access Saiki via API endpoints.

Added support for streamable HTTP in MCP servers

Added support for Saiki agents to be used via MCP


rahulkarajgikar:
@CodiumAI-Agent /review

src/api/server.ts
```diff
@@ -0,0 +1,156 @@
...
+        });
+        ws.on('close', () => {
+            logger.info('WebSocket client disconnected.');
+        });
+        ws.on('error', (error) => {
+            logger.error(`WebSocket error: ${error.message}`);
+        });
+    });
+
+    // --- Expose Saiki via MCP protocol on /mcp ---
```
rahulkarajgikar: nice!

src/api/server.ts
```diff
@@ -0,0 +1,156 @@
...
+            const text = await llmService.completeTask(message);
+            return { content: [{ type: 'text', text }] };
+        }
+    );
+    // Initialize the MCP protocol on this transport before mounting endpoints
+    logger.info('Initializing MCP protocol server...');
+    await mcpServer.connect(mcpTransport);
+    logger.info('✅ MCP server protocol initialized on /mcp');
+    // Mount /mcp for JSON-RPC and SSE handling (so baseUrl is /mcp)
+    app.post('/mcp', express.json(), (req, res) => {
```
rahulkarajgikar, line 168: have we tested that this works? 

also i'm assuming that with this, users can connect saiki as an MCP server by hosting saiki at a URL, is this right?

src/api/websocket-subscriber.ts
```diff
@@ -0,0 +1,47 @@
+import { EventEmitter } from 'events';
+import { WebSocketServer, WebSocket } from 'ws';
+import { EventSubscriber } from './types.js';
+
+/**
+ * Subscribes to core events and broadcasts them over all open WebSocket connections.
+ */
+export class WebSocketEventSubscriber implements EventSubscriber {
```
rahulkarajgikar, line 8: this is good, we should refactor `CLISubscriber` to also use this interface - would make `cli.ts` code simpler

we just need to add the `subscribe` method and move all of the actual subscription logic like this:

old code:
```
        logger.info('Setting up CLI event subscriptions...');
        const cliSubscriber = new CLISubscriber();
        agentEventBus.on('llmservice:thinking', cliSubscriber.onThinking.bind(cliSubscriber));
        agentEventBus.on('llmservice:chunk', cliSubscriber.onChunk.bind(cliSubscriber));
        agentEventBus.on('llmservice:toolCall', cliSubscriber.onToolCall.bind(cliSubscriber));
        agentEventBus.on('llmservice:toolResult', cliSubscriber.onToolResult.bind(cliSubscriber));
        agentEventBus.on('llmservice:response', cliSubscriber.onResponse.bind(cliSubscriber));
        agentEventBus.on('llmservice:error', cliSubscriber.onError.bind(cliSubscriber));
        agentEventBus.on(
            'llmservice:conversationReset',
            cliSubscriber.onConversationReset.bind(cliSubscriber)
        );
```

new code:

```
    const cliSubscriber = new CLISubscriber();
    logger.info('Setting up CLI event subscriptions...');
    cliSubscriber.subscribe(agentEventBus);
```

src/client/manager.ts
```diff
@@ -256,6 +256,13 @@ export class MCPClientManager {
             logger.info(`Successfully connected and cached new server '${name}'`);
         } catch (error) {
             const errorMsg = error instanceof Error ? error.message : String(error);
+            // If server has already been initialized (duplicate handshake), treat as connected
+            if (errorMsg.includes('Server already initialized')) {
```
rahulkarajgikar: this is very brittle, in general it's bad to have substring/text based checks on errors - if the error message changes then the code breaks

better to either have a custom exception or just drop this code completely - we already do have a similar check in lines 245-247 by checking our internal registry

is there any specific reason we added this code?

src/client/mcp-client.ts
```diff
@@ -187,6 +191,38 @@ export class MCPClient implements IMCPClient {
         }
     }
 
+    /**
+     * Connect to an MCP server via Streamable HTTP transport
+     */
+    private async connectViaHttp(
+        baseUrl: string,
+        headers: Record<string, string>,
+        timeout: number | undefined,
```
rahulkarajgikar: timeout is unused, we can remove this argument

currently `this.timeout` is used for request timeouts

we could actually have a separate connection timeout for initializing the server connection, we can do in a separate PR

src/client/mcp-client.ts
```diff
@@ -256,26 +292,26 @@ export class MCPClient implements IMCPClient {
             const listToolResult = await this.client.listTools({});
             logger.debug(`listTools result: ${JSON.stringify(listToolResult, null, 2)}`);
 
-            // Assume listToolResult.tools exists and iterate
+            // Populate tools
             if (listToolResult && listToolResult.tools) {
                 listToolResult.tools.forEach((tool: any) => {
                     if (!tool.description) {
-                        throw new Error(`Tool '${tool.name}' is missing a description`);
+                        logger.warn(`Tool '${tool.name}' is missing a description`);
```
rahulkarajgikar, line 308: i don't think we should change this - we should enforce tools have descriptions unless there's a good reason not to. this would lead to saiki silently skipping tools which is much harder to notice and debug

src/client/mcp-client.ts
```diff
@@ -256,26 +292,26 @@ export class MCPClient implements IMCPClient {
...
                     }
                     if (!tool.inputSchema) {
                         throw new Error(`Tool '${tool.name}' is missing an input schema`);
                     }
                     tools[tool.name] = {
-                        description: tool.description,
+                        description: tool.description ?? '',
```
rahulkarajgikar, line 314: same here, no default

src/client/mcp-client.ts
```diff
@@ -256,26 +292,26 @@ export class MCPClient implements IMCPClient {
...
                     };
                 });
             } else {
                 throw new Error('listTools did not return the expected structure: missing tools');
             }
         } catch (error) {
-            logger.error('Failed to get tools from MCP server:', error);
-            throw error;
+            logger.warn('Failed to get tools from MCP server, proceeding with zero tools:', error);
+            return tools;
```
rahulkarajgikar: same here

src/api/server.ts
```diff
@@ -0,0 +1,156 @@
...
+    const server = http.createServer(app);
+    const wss = new WebSocketServer({ server });
+
+    // set up event broadcasting over WebSocket
+    const webSubscriber = new WebSocketEventSubscriber(wss);
+    logger.info('Setting up API event subscriptions...');
+    webSubscriber.subscribe(agentEventBus);
+
+    // HTTP endpoints
+    app.post('/api/message', express.json(), async (req, res) => {
```
rahulkarajgikar, line 30: we should rethink API names, but we can defer this for later. let's add a TODO comment at the top of this function that API names are work in progress and might be refactored/renamed

rahulkarajgikar (changes requested):
mostly looks good, few minor comments

shaunak99:
To use an agent built with saiki as an MCP server, we will setup 2 config files for 2 separate agents. One will act as the client, the other will act as the remote MCP server. 

To demonstrate locally, first startup the remote instance of saiki on a port using
```
saiki --mode web --web-port 3001 --config-file .\configuration\remote_agent_config.yml
```

Final Output:
```
INFO: WebUI server started successfully. Accessible at: http://localhost:3001 ...
```

Then configure your client config with the MCP config for this agent:
```
mcpServers:
  agent:
    type: http
    baseUrl: http://localhost:3001/mcp
```

Sample interaction between client and remote agent:
image

src/client/mcp-client.ts
shaunak99: Some servers may expose only resources/prompts without tools.

src/client/mcp-client.ts
shaunak99: refer above

src/client/mcp-client.ts
shaunak99: Made this change primarily because some of the public MCPs i tried did not have description but they worked with the just the toolname and inputSchema. Description is a good to have but not necessary param so I thought a warning may be apt; any better way to handle it?

src/client/manager.ts
shaunak99: Good catch, name duplication check is already present in the beginning

src/api/server.ts
shaunak99: Yes! Saiki itself becomes an MCP. Have tested it, will add a comment with a test example to illustrate

app/cli/cli-subscriber.ts
```diff
@@ -14,10 +16,20 @@ import chalk from 'chalk';
...
     private currentLines: number = 0;
 
+    subscribe(eventBus: EventEmitter): void {
+        eventBus.on('llmservice:thinking', this.onThinking.bind(this));
+        eventBus.on('llmservice:chunk', this.onChunk.bind(this));
+        eventBus.on('llmservice:toolCall', this.onToolCall.bind(this));
+        eventBus.on('llmservice:toolResult', this.onToolResult.bind(this));
+        eventBus.on('llmservice:response', this.onResponse.bind(this));
+        eventBus.on('llmservice:error', this.onError.bind(this));
+        eventBus.on('llmservice:conversationReset', this.onConversationReset.bind(this));
```
rahulkarajgikar, line 30: nice, this is so much better

app/web/server.ts
```diff
@@ -1,166 +1,43 @@
...
-                        })
-                    );
+    server.listen(port, '0.0.0.0', () => {
+        const networkInterfaces = os.networkInterfaces();
+        let localIp = 'localhost';
+        Object.keys(networkInterfaces).forEach(ifaceName => {
+            networkInterfaces[ifaceName]?.forEach(iface => {
+                if (iface.family === 'IPv4' && !iface.internal) {
+                    localIp = iface.address;
+                    return;
```
rahulkarajgikar: what does this do?

src/client/mcp-client.ts
rahulkarajgikar: I see, I guess then we can keep the warning message for now and skip silently, let's see if MCP enforces this in their own SDK

src/client/mcp-client.ts
rahulkarajgikar: Got it, also for logger messages, you have to use the other quotes (tilted ones, can't type on phone), and keep the entire object in one quote without commas. Refer some other logs in other files - comma separated values don't work with Winston logging library so whatever goes after the comma won't show up in the logs
 

The object to be logged would look something like json.stringify(object, null, 2) (don't recall exact syntax)

rahulkarajgikar: so we can make this change in all the logs we've added in this PR (and any others if you see them)

rahulkarajgikar (changes requested):
nice! Looks good after refactors, just a few minor comments then we're good to merge

rahulkarajgikar:
> To use an agent built with saiki as an MCP server, we will setup 2 config files for 2 separate agents. One will act as the client, the other will act as the remote MCP server. 
> 
> 
> 
> To demonstrate locally, first startup the remote instance of saiki on a port using
> 
> ```
> 
> saiki --mode web --web-port 3001 --config-file .\configuration\remote_agent_config.yml
> 
> ```
> 
> 
> 
> Final Output:
> 
> ```
> 
> INFO: WebUI server started successfully. Accessible at: http://localhost:3001 ...
> 
> ```
> 
> 
> 
> Then configure your client config with the MCP config for this agent:
> 
> ```
> 
> mcpServers:
> 
>   agent:
> 
>     type: http
> 
>     baseUrl: http://localhost:3001/mcp
> 
> ```
> 
> 
> 
> Sample interaction between client and remote agent:
> 
> image
> 
> 

let's also add this example into our documentation, I'll include it in docs PR, commenting so we remember

app/web/server.ts
shaunak99: This creates a URL that is available on LAN so you can connect to agents from other devices also on the LAN

shaunak99:
Will add the the docs for this in the next PR with a few more examples

rahulkarajgikar (approved):

src/api/server.ts
```diff
@@ -0,0 +1,161 @@
...
+            return { content: [{ type: 'text', text }] };
+        }
+    );
+    // Initialize the MCP protocol on this transport before mounting endpoints
+    logger.info('Initializing MCP protocol server...');
+    await mcpServer.connect(mcpTransport);
+    logger.info('✅ MCP server protocol initialized on /mcp');
+    // Mount /mcp for JSON-RPC and SSE handling (so baseUrl is /mcp)
+    app.post('/mcp', express.json(), (req, res) => {
+        mcpTransport.handleRequest(req, res, req.body).catch(err => logger.error('MCP POST error:', err));
```
rahulkarajgikar: nit: fix this log

src/api/server.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+    app.get('/mcp', (req, res) => {
+        mcpTransport.handleRequest(req, res).catch(err => logger.error('MCP GET error:', err));
```
rahulkarajgikar: nit: fix this log

src/api/server.ts
shaunak99: my b

PR #96, opened 2025-05-09
Enhancing server/api layer & added new telegram and discord adapters

shaunak99:
This PR introduces enhancements to the api layer by enhancing the mcp handler for interacting with Saiki as an MCP server. We have introduced the concept of AgentCard from A2A and implemented an endpoint to fetch the AgentCard using a http `GET` request as `https://DOMAIN/.well-known/agent.json` as per the A2A specs. Saiki also exposes this AgentCard via an MCP resource. 

To make agent-2-agent communication feasible between 2 saiki instances, we have also implemented a resource contributor to the systemPrompt contributor.  This fetches the AgentCard of any MCP-connected saiki agent, say `saiki-remote`, to the context of the client side agent, say `saiki-host` and can be used by the host/client application to interact with the remote agent.

Additionally, we have also added support for new platforms interfaces, namely telegram & discord. This allows users to spin up new Saiki agents that you can interact with via Telegram and Discord.

## Summary by CodeRabbit

- **New Features**
  - Added Discord and Telegram bot integrations for conversational AI with support for text, images, typing indicators, and tool call notifications.
  - Introduced dynamic resource data prompts to enrich AI responses with real-time resource information.
  - Implemented agent card support compliant with the A2A specification, including a public endpoint serving agent metadata.
- **Improvements**
  - Enhanced system prompt functionality with new dynamic contributors.
  - Modularized and parameterized MCP server and agent card initialization with validation.
  - Extended CLI to support new run modes for Discord and Telegram bots with environment variable validations.
  - Renamed startup functions for CLI and WebUI for consistency.
  - Expanded tool confirmation provider to support Discord and Telegram modes.
- **Documentation**
  - Added detailed README files for Discord and Telegram bot setup and usage.
  - Updated main README with instructions for running Saiki in bot modes.
- **Dependencies**
  - Added `discord.js` and `node-telegram-bot-api` packages.


app/index.ts
```diff
@@ -109,6 +111,7 @@ const normalizedConfigPath = resolvePackagePath(configFile, resolveFromPackageRo
 
 // basic validation of options here
 try {
+    // Allow discord and telegram modes to bypass certain CLI-specific validations if necessary
```
rahulkarajgikar: remove this comment

app/index.ts
```diff
@@ -139,9 +142,17 @@ async function startAgent() {
...
     let services;
     try {
+        // Determine the runMode for service initialization to satisfy ToolConfirmationProvider
+        // Currently bots are run on top of web server, so we need to pass web mode to service initializer
+        // TODO: Update with a NoOp runMode for actual bot
+        let serviceInitRunMode = runMode;
+        if (runMode === 'discord' || runMode === 'telegram') {
+            serviceInitRunMode = 'web'; // Use 'web' (NoOp) for service init when actual mode is a bot
+        }
+
```
rahulkarajgikar: let's not do this, just pass the runMode all the way down

runMode is used in `createToolConfirmationProvider`

```
export function createToolConfirmationProvider(runMode: 'cli' | 'web'): ToolConfirmationProvider {
    switch (runMode) {
        case 'cli':
            return new CLIConfirmationProvider();
        case 'web':
            // Fallback: No-op provider for now. Replace with real provider when available.
            return new NoOpConfirmationProvider();
        default:
            throw new Error(`Unknown run mode for ToolConfirmationProvider: ${runMode}`);
    }
}
```

we can update here to add new cases for discord and telegram instead, easier to follow this

app/index.ts
```diff
@@ -139,9 +142,17 @@ async function startAgent() {
...
         services = await createAgentServices(normalizedConfigPath, cliArgs, {
             connectionMode,
-            runMode,
+            runMode: serviceInitRunMode, // Pass the adjusted runMode here
```
rahulkarajgikar: then we can just directly use runMode as we were using before

app/index.ts
```diff
@@ -169,9 +180,16 @@ async function startAgent() {
...
     } else if (runMode === 'web') {
-        // Run WebUI
-        initializeWebUI(clientManager, llmService, agentEventBus, webPort);
+        // Run WebUI with configured MCP identity (pass agentCard only)
+        const agentCard = services.configManager.getConfig().agentCard ?? {};
+        initializeWebUI(clientManager, llmService, agentEventBus, webPort, agentCard);
         logger.info(`WebUI available at http://localhost:${webPort}`, null, 'magenta');
+    } else if (runMode === 'discord') {
+        logger.info('Starting Discord bot...', null, 'cyanBright');
+        await startDiscordBot(normalizedConfigPath);
```
rahulkarajgikar: there is inconsistency here in how we are handling dependencies for discord/telegram vs cli/webUI

CLI/WebUI - we are initializing the services first and then passing them in as dependencies
discord/telegram - we are first initializing the services, then we pass in the config file, then we again initialize their fresh set of services

we should keep this more consistent.

firstly we should rename `initializeWebUI`, `runHeadlessCli`, `runAiCli` to have the same prefix `start` so that the code is more consistent across the different modes.

now for the dependencies

Option 1: Every `start` function/mode accepts the already initialized services - initialize services once - pass them into required function. all functions will have similar signature

we can pass the services here as well instead of the config path for both `startDiscordBot` and `startTelegramBot`, similar to how we are doing in `initializeWebUI`

right now for discord and telegram we are doing the service initialization twice.
by doing this, we don't need to reconfigure the agent services in `startTelegramBot` and `startDiscordBot`

pro: we avoid code duplication
con: for some specific file paths/env variables we will need to have another if-else tree before we hit this function (like the TELEGRAM_FILE_PATH env variable)


Option 2: Every `start` function/mode accepts just the config file path and creates its own dependencies

this would have some duplicated code
For this, we would remove the `createAgentServices` block from above and handle this in the respective `start` functions

pro: each mode `owns` its own logic for initializing the services and any env variables/overrides
con: code duplication 


I'm ok with either approach as long as we are consistent, we can discuss which one we prefer

I am leaning towards option 2 - this allows us to have separate env variables for the file path for each mode. [and potentially other mode specific logic]. Even though there might be some code duplication, this separation allows for more flexibility.

src/ai/systemPrompt/manager.ts
```diff
@@ -27,6 +27,7 @@ export class PromptManager {
         // Inline loader logic from loader.ts:
         const defaultContributors: ContributorConfig[] = [
             { id: 'dateTime', type: 'dynamic', priority: 10, source: 'dateTime', enabled: true },
+            { id: 'resources', type: 'dynamic', priority: 20, source: 'resources', enabled: true },
```
rahulkarajgikar: should we make this a default? this could cause bloat in the system prompt if there are many mcp servers and we don't need all the resources

src/api/mcp_handler.ts
```diff
@@ -0,0 +1,91 @@
...
+        { name: agentName, version: agentVersion },
+        {
+            capabilities: {
+                resources: {}, // Declare resource capability
+            },
+        }
+    );
+
+    // Register the primary 'chat' tool with fixed details
+    const toolName = 'chat_with_agent'; // Simplified tool name
```
rahulkarajgikar, line 32: chat_with_saiki!

src/api/mcp_handler.ts
```diff
@@ -0,0 +1,91 @@
...
+    const toolDescription = 'Allows you to chat with the an AI agent. Send a message to interact.';
```
rahulkarajgikar, line 33: chat with Saiki, the AI Agent

src/api/server.ts
```diff
@@ -96,7 +101,7 @@ export async function initializeApi(
 
     // WebSocket handling
     // handle inbound client messages over WebSocket
-    wss.on('connection', (ws: WebSocket) => {
+    wss.on('connection', (ws) => {
```
rahulkarajgikar: keep the type

src/config/schemas.ts
```diff
@@ -54,3 +54,44 @@ export const llmConfigSchema = z
     });
 
 // You can add more schemas for AgentConfig, etc., as needed.
+
+export const AgentCardOverrideSchema = z
```
rahulkarajgikar: so i'm guessing ideally we would not need this schema internally and we would fetch this from A2A sdk?

src/config/types.ts
```diff
@@ -105,3 +105,36 @@ export type Source = 'file' | 'cli' | 'default';
  * Provenance for CLI-overridable LLM fields only
  */
 export type LLMProvenance = Record<LLMOverrideKey, Source>;
+
+// Agent Card interface (based on A2A specification)
+export interface AgentCard {
```
rahulkarajgikar: can we just use the interface we made earlier instead? this seems like a duplicate interface

we can just do this and not need to define a new interface
`export interface AgentCard = z.infer<typeof AgentCardSchema>`

also yeah i'm realizing a lot of the schemas we have are duplicated and have this same problem

the whole `SchemaFromInterface` is unnecessary because i was trying to create zod schemas using interfaces as the source of truth.

what we actually should have done is use the zod schema as the source of truth and generate the interfaces using z.infer like we've done for `AgentCardOverride` - this is much simpler.

 i will update the other schemas as part of a separate PR

src/api/server.ts
```diff
@@ -143,38 +148,92 @@ export async function initializeApi(
...
+            credentials: cardConfig.authentication?.credentials,
+        },
+        defaultInputModes: cardConfig.defaultInputModes ?? ['application/json', 'text/plain'],
+        defaultOutputModes: cardConfig.defaultOutputModes ?? [
+            'application/json',
+            'text/event-stream',
+            'text/plain',
+        ],
+        skills: (cardConfig.skills ?? []) as AgentCard['skills'], // Skills will be populated based on registered MCP tools for the A2A card
+    };
```
rahulkarajgikar: is all this necessary? can we just use the cardConfig directly?

src/api/server.ts
```diff
@@ -1,21 +1,26 @@
...
+import type { AgentCardOverride } from '../config/schemas.js';
+import { ZodError } from 'zod';
 
 // TODO: API endpoint names are work in progress and might be refactored/renamed in future versions
 export async function initializeApi(
     clientManager: MCPClientManager,
     llmService: ILLMService,
-    agentEventBus: EventEmitter
+    agentEventBus: EventEmitter,
+    agentCardOverride?: Partial<AgentCard>
```
rahulkarajgikar, line 22: do we need this override?

rahulkarajgikar (changes requested):

app/index.ts
shaunak99: Makes sense. 

I don't know if this falls under option 1 but ideally we should treat all modes as just an interface layer on top of a single instance of createAgentServices. This would allow us to have the flexibility of multiple ways to interact with the same instance of the agent, but across interfaces. 

Telegram and Discord would just be like the WebUI, a layer on top of the `initializeWebUI` or equivalent. Use case - this would enable building a cross-platform experience over the same agent instance. 

For now I am implementing Option 2, we can revisit this in a future PR

src/ai/systemPrompt/manager.ts
shaunak99: changing this to `enables: false`

src/api/mcp_handler.ts
shaunak99: skipping as it will conflict with systemprompt, leaving it to be generic

src/api/mcp_handler.ts
shaunak99: skipping as it will conflict with any custom systemprompt, leaving it generic

src/config/schemas.ts
shaunak99: Yes, implemented the same spec

src/api/server.ts
shaunak99: Wanted to have a default cardConfig in the code somewhere, have added a suggested refactor to move it out of the api layer. Do share your thoughts

src/api/server.ts
shaunak99: Coderabbit suggested - https://github.com/truffle-ai/saiki/pull/96#discussion_r2081248867

Let me know if you want to keep or remove

app/index.ts
rahulkarajgikar: interesting idea - we can add this as a todo comment at the top of this function as a thought to revisit this later so we remember. that could be a useful use-case

src/api/server.ts
rahulkarajgikar: looks like coderabbit mentioned we should have validations - didn't see it suggest the override

if we aren't using the override anywhere, let's just remove it - can always add something like this later if we do need it

app/discord/bot.ts
```diff
@@ -0,0 +1,111 @@
...
+                await message.channel.sendTyping();
+                const responseText = await llmService.completeTask(userText, imageDataInput);
+                await message.reply(responseText);
+            } catch (error) {
+                console.error('Error handling Discord message', error);
+                await message.reply(`Error: ${error.message}`);
+            } finally {
+                agentEventBus.off('llmservice:toolCall', toolCallHandler);
+            }
+        }
```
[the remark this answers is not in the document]
rahulkarajgikar: interesting idea, let's actually do the simplest form of rate limiting we can driven through ENV variable for configurability

app/discord/bot.ts
```diff
@@ -0,0 +1,111 @@
...
+                message.channel.send(
+                    `⚙️ Calling tool **${toolName}** with args: ${JSON.stringify(args)}`
+                );
+            };
+            agentEventBus.on('llmservice:toolCall', toolCallHandler);
+
+            try {
+                await message.channel.sendTyping();
+                const responseText = await llmService.completeTask(userText, imageDataInput);
+                await message.reply(responseText);
```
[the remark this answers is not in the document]
rahulkarajgikar: good catch lil bot

app/telegram/bot.ts
```diff
@@ -0,0 +1,160 @@
...
+
+        if (!userText) return;
+
+        // Subscribe for toolCall events
+        const toolCallHandler = (toolName: string, args: any) => {
+            bot.sendMessage(chatId, `Calling *${toolName}* with args: ${JSON.stringify(args)}`, {
+                parse_mode: 'Markdown',
+            });
+        };
+        agentEventBus.on('llmservice:toolCall', toolCallHandler);
```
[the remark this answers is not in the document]
rahulkarajgikar: bro really loves his error handling, let's add this

app/telegram/bot.ts
```diff
@@ -0,0 +1,160 @@
...
+        const chatId = callbackQuery.message?.chat.id;
+        if (!chatId) return;
+        if (action === 'reset') {
+            await llmService.resetConversation();
+            await bot.sendMessage(chatId, '🔄 Conversation has been reset.');
+        } else if (action === 'help') {
+            await bot.sendMessage(chatId, 'Send me text or images and I will respond.');
+        }
+        await bot.answerCallbackQuery(callbackQuery.id);
+    });
```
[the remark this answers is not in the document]
rahulkarajgikar: good suggestion to notify user if there is a failure, good UX

src/api/server.ts
shaunak99: Using it here for injecting agentCard from config
```
    // Apply agentCard overrides (if any)
    const rawOverride = agentCardOverride ?? {};
    let cardConfig: AgentCardOverride;
    try {
        cardConfig = AgentCardOverrideSchema.parse(rawOverride);
    } catch (err: unknown) {
        if (err instanceof ZodError) {
            logger.error('Invalid agentCard override:');
            err.errors.forEach((e) => {
                const path = e.path.join('.') || '<root>';
                logger.error(`- ${path}: ${e.message}`);
            });
        } else {
            logger.error('Invalid agentCard override:', err);
        }
        process.exit(1);
    }
```

app/discord/bot.ts
```diff
@@ -0,0 +1,178 @@
...
+
+        // Handle attachments (images)
+        if (message.attachments.size > 0) {
+            const attachment = message.attachments.first();
+            if (attachment && attachment.url) {
+                const { base64, mimeType } = await downloadFileAsBase64(attachment.url);
+                imageDataInput = { image: base64, mimeType };
+                userText = message.content || '';
+            }
+        }
```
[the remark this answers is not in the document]
shaunak99: Will do this in the future if needed to make this more robust

README.md
```diff
@@ -69,6 +69,22 @@ npm start -- --mode web --web-port 3000
...
+```bash
+saiki --mode discord
+```
+Make sure you have `DISCORD_BOT_TOKEN` set in your environment. See [here](app/discord/README.md) for more details.
+
+**Telegram Bot:**
+```bash
+saiki --mode telegram
+```
+Make sure you have `TELEGRAM_BOT_TOKEN` set in your environment. See [here](app/telegram/README.md) for more details.
```
rahulkarajgikar, line 86: nice - short and sweet

src/api/server.ts
rahulkarajgikar: ok cool we can keep for now

rahulkarajgikar (approved):
looks good to me, let's just ensure to address lil bots comments for the important things (memory leaks/size limits/ basic rate limiting, etc.)

src/api/server.ts
rahulkarajgikar: maybe setting the defaults within the zodSchema itself is better.

that way, when we override the defaults like `AgentCardSchema.parse(overrides)` it can set defaults for everything that doesn't have an override maybe.

essentially - zod schema owns the default and parsing. we can update the overrides from the cardConfig

so the code becomes:
1. AgentCardSchema - this is a zod schema that defines shapes and defaults
2. AgentCard - type inferred from schema for use in the code
3. code above then gets overrides as shape AgentCard from the arguments.
4. call AgentCardSchema.parse(overrides) - this will set overrides for anything that exists, and use defaults otherwise. So we have set both the defaults + the overrides
5. we can even wrap this parsing overrides into a separate function that can be made into an API to enable hot reloading the agent card

PR #106, opened 2025-05-16
New WebUI using NextJS

shaunak99:
The old webui has been replaced with a fresh NextJS webui. This new webui allows viewing MCP servers, connecting new servers, and also an MCP playground to test out functionality of individual MCP servers.

## Summary by CodeRabbit

- **New Features**
  - Introduced a modern web UI built with Next.js, featuring a chat interface, server management panel, and configuration export.
  - Added a Playground interface for managing servers and executing tools with live input validation and rich result rendering.
  - Implemented REST API endpoints for server listing, tool discovery, tool execution, and configuration export in YAML format.
  - Enabled users to connect new servers via a guided modal dialog.

- **Bug Fixes**
  - Improved error handling and feedback throughout the UI for server connections, tool execution, and configuration export.

- **Style**
  - Added comprehensive theming with light/dark mode, custom fonts, and responsive layouts using Tailwind CSS.

- **Documentation**
  - Added README and configuration files for easier development, deployment, and usage guidance.

- **Chores**
  - Set up project tooling including ESLint, TypeScript, Tailwind CSS, and package management for both backend and frontend.
  - Updated lint scripts to cover more file types and integrated Next.js-specific linting.


src/app/api/server.ts
```diff
@@ -175,5 +225,27 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+    // Export current AgentConfig as YAML, omitting sensitive fields
+    app.get('/api/config.yaml', async (req, res) => {
+        try {
+            // Deep clone and sanitize
+            const rawConfig = agent.configManager.getConfig();
+            const exportConfig = JSON.parse(JSON.stringify(rawConfig));
+            // Remove sensitive API key
+            if (exportConfig.llm && 'apiKey' in exportConfig.llm) {
+                exportConfig.llm.apiKey = 'SET_YOUR_API_KEY_HERE';
+            }
```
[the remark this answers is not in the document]
rahulkarajgikar: we don't have `exportConfig.credentials` at the moment but this is an interesting thought to separate out credentials in a different section of config. 

was thinking mainly we can use env variables for api keys

src/app/index.ts
```diff
@@ -175,8 +177,29 @@ async function startApp() {
...
+        );
+        spawn('npm', ['run', 'dev', '--', '--port', String(frontPort)], {
+            cwd: nextCwd,
+            shell: true,
+            stdio: 'inherit',
+            env: {
+                ...process.env,
+                NODE_ENV: 'development',
+                API_PORT: String(apiPort),
+                NEXT_PUBLIC_WS_URL: `ws://localhost:${apiPort}`,
```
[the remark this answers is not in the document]
rahulkarajgikar: this makes sense, useful addition to allow changing the URL path which is helpful for devs

we could eventually expose this as a CLI argument to allow users to specify URL path maybe? or just passing through env variable enables docker based builds or hosting on vercel/netlify

src/app/webui/components/ChatApp.tsx
```diff
@@ -0,0 +1,118 @@
...
+      link.download = `${exportName}.yml`;
+      document.body.appendChild(link);
+      link.click();
+      document.body.removeChild(link);
+      URL.revokeObjectURL(url);
+      setExportOpen(false);
+    } catch (err) {
+      console.error('Export failed:', err);
+    }
+  }, [exportName]);
```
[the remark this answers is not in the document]
rahulkarajgikar: useful

src/app/webui/components/ConnectServerModal.tsx
```diff
@@ -0,0 +1,219 @@
...
+    setError(null);
+    setIsSubmitting(true);
+
+    if (!serverName.trim()) {
+      setError('Server name is required.');
+      setIsSubmitting(false);
+      return;
+    }
+
+    let config: Record<string, any> = { type: serverType };
```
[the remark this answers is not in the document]
rahulkarajgikar: there is an McpServerConfig type in core, that is probably better to use here

src/app/webui/components/MessageList.tsx
```diff
@@ -0,0 +1,204 @@
...
+    );
+  }
+
+  return (
+    <div id="message-list-container" className="flex flex-col space-y-3 px-4 py-2">
+      {messages.map((msg, idx) => {
+        const msgKey = msg.id ?? `msg-${idx}`;
+        const isUser = msg.role === 'user';
+        const isAi = msg.role === 'ai';
+        const isSystem = msg.role === 'system';
```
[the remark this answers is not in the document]
rahulkarajgikar: we are using the 'assistant' role everywhere in the backend - it probably makes sense to stick to one name everywhere. i'm fine with either `ai` or `assistant` as long as its the same in both backend and here

src/app/webui/components/ServersPanel.tsx
```diff
@@ -0,0 +1,305 @@
...
+      } else {
+        console.log("No MCP servers found or returned from API.");
+      }
+    } catch (err: any) {
+      handleError(err.message, 'servers');
+    } finally {
+      setIsLoadingServers(false);
+    }
+  }, []);
+
```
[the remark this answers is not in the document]
rahulkarajgikar: good catch

src/app/webui/components/hooks/useChat.ts
```diff
@@ -0,0 +1,210 @@
...
+            `msg-${Date.now()}-${Math.random().toString(36).substring(2, 9)}`;
+
+        wsRef.current = ws;
+        ws.onopen = () => setStatus('open');
+        ws.onclose = () => setStatus('closed');
+        ws.onmessage = (event: MessageEvent) => {
+            const msg = JSON.parse(event.data);
+            const payload = msg.data || {};
+            switch (msg.event) {
+                case 'thinking':
```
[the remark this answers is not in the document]
rahulkarajgikar: makes sense

src/app/webui/components/ui/dialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+import * as DialogPrimitive from "@radix-ui/react-dialog"
+import { XIcon } from "lucide-react"
+
+import { cn } from "@/lib/utils"
+
+function Dialog({
+  ...props
+}: React.ComponentProps<typeof DialogPrimitive.Root>) {
+  return <DialogPrimitive.Root data-slot="dialog" {...props} />
+}
```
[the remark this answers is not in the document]
rahulkarajgikar: #honestly no idea what this means lol

src/app/api/server.ts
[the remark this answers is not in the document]
shaunak99: Not needed for now

src/app/webui/components/ui/select.tsx
```diff
@@ -0,0 +1,185 @@
...
+              "h-[var(--radix-select-trigger-height)] w-full min-w-[var(--radix-select-trigger-width)] scroll-my-1"
+          )}
+        >
+          {children}
+        </SelectPrimitive.Viewport>
+        <SelectScrollDownButton />
+      </SelectPrimitive.Content>
+    </SelectPrimitive.Portal>
+  )
+}
```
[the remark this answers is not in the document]
shaunak99: predefined component, lets not edit

src/app/webui/components/ui/dialog.tsx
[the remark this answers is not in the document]
shaunak99: this is a pre-built component let it be as is

src/app/index.ts
```diff
@@ -180,8 +183,47 @@ async function startApp() {
...
+        );
+        const nextProc = spawn('npm', ['run', 'dev', '--', '--port', String(frontPort)], {
+            cwd: nextCwd,
+            shell: true,
+            stdio: 'inherit',
+            env: {
+                ...process.env,
+                NODE_ENV: 'development',
+                API_PORT: String(apiPort),
+                NEXT_PUBLIC_WS_URL:
```
rahulkarajgikar, line 206: @coderabbitai explain this block

src/app/webui/components/hooks/useChat.ts
```diff
@@ -0,0 +1,249 @@
...
+    // Content for text-based messages
+    content: string | null | Array<TextPart | ImagePart>;
+    // Optional image data sent directly in user messages
+    imageData?: { base64: string; mimeType: string };
+    // Tool call and result metadata
+    toolName?: string;
+    toolArgs?: any;
+    toolResult?: any;
+}
+
```
rahulkarajgikar: can we re-use these types from core

@coderabbitai check if we have types like this anywhere in core and if we can/should use those

src/app/webui/eslint.config.mjs
```diff
@@ -0,0 +1,16 @@
...
+
+const compat = new FlatCompat({
+  baseDirectory: __dirname,
+});
+
+const eslintConfig = [
+  ...compat.extends("next/core-web-vitals", "next/typescript"),
+];
+
+export default eslintConfig;
```
rahulkarajgikar: @coderabbitai how does this eslint.config.mjs work with the eslint.config.js in the root folder? will they conflict or work together? and what do you recommend (read the root file as well and tell me)

src/app/webui/next.config.ts
```diff
@@ -0,0 +1,38 @@
...
+            allowedOrigins.push(`http://${iface.address}:3000`);
+        }
+    })
+);
+const nextConfig: NextConfig = {
+    reactStrictMode: true,
+    // Allow static asset requests from these origins in dev mode
+    allowedDevOrigins: allowedOrigins,
+    async rewrites() {
+        const apiPort = process.env.API_PORT ?? '3001';
```
rahulkarajgikar, line 20: we should update src/app/index.ts to also support `process.env.API_PORT` otherwise if this env variable is passed it wont work correctly

src/app/webui/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+        "name": "next"
+      }
+    ],
+    "paths": {
+      "@/*": ["./*"]
+    }
+  },
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "exclude": ["node_modules"]
+}
```
rahulkarajgikar: @coderabbitai is this consistent with the tsconfig.json at the root of the project?

please read both files and suggest changes, if any

src/app/api/server.ts
```diff
@@ -79,6 +80,14 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
         }
         try {
             await agent.connectMcpServer(name, config);
+            // Add dynamic server config to in-memory AgentConfig
+            try {
+                const cfg = agent.configManager.getConfig();
+                if (!cfg.mcpServers) cfg.mcpServers = {} as any;
+                cfg.mcpServers[name] = config;
```
rahulkarajgikar, line 89: is this cfg used anywhere? doesn't seem like we are using it

src/app/api/server.ts
rahulkarajgikar, line 269: i don't think this will fetch the updated config after user changes prompts/model etc, because i don't think we are actually updating the base `agent.configManager.config` when those things change

can revisit this logic later and see if we should actually update the base config in core or if we should just store the updated state in frontend

src/app/webui/components/hooks/useChat.ts
rahulkarajgikar: yes let's do this so that the types stay in sync

also update the import to use @core instead of ../../.. relative import for modularity

src/app/webui/eslint.config.mjs
rahulkarajgikar: let's go with option 1

with option 1:
1. would we get rid of the eslint.config.js in the app folder?
2. can you suggest the exact edit/prompt to use

src/app/api/webui.ts
```diff
@@ -6,32 +6,28 @@ import os from 'os';
 import type { AgentCard } from '@core/index.js';
 import { SaikiAgent } from '@core/index.js';
 
-export async function startWebUI(
+export async function startServer(
```
rahulkarajgikar: nit: add a docstring explaining that the apis are hosted on an express server and frontend is hosted on a next js server and that this function is responsible for starting both of these servers

could rename to startApiAndWebUIServers

rahulkarajgikar: wait nvm this is just for the express server

then we can rename this to just `startApiServer` - `startServer` is vague

src/app/index.ts
```diff
@@ -180,8 +183,47 @@ async function startApp() {
...
+        nextProc.on('error', (err) => {
+            logger.error('Failed to start Next.js process:', err);
+        });
+        nextProc.on('exit', (code, signal) => {
+            logger.info(`Next.js process exited with code ${code} and signal ${signal}`);
+        });
+        // Start Express API server (for WebSocket and HTTP endpoints)
+        startServer(agent, apiPort, agentCard);
+        logger.info(`API endpoints available at http://localhost:${apiPort}`, null, 'magenta');
+        logger.info(`Frontend available at http://localhost:${frontPort}`, null, 'magenta');
```
rahulkarajgikar: we aren't consistently sticking to a convention of env variable or localhost

this will be confusing if `process.env.NEXT_PUBLIC_WS_URL` is passed - it will still log localhost

similar for the `process.env.API_PORT` we added

we should fix this - essentially standardize on a few env variables controlling express and nextjs servers and then use them everywhere properly. ok to do this in a future PR as well but yeah we should do it

src/app/webui/app/api/config/route.ts
```diff
@@ -0,0 +1,30 @@
...
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify(body),
+        });
+        const data = await res.json();
+        return NextResponse.json(data, { status: res.status });
+    } catch (err) {
+        console.error('Error proxying POST /api/config:', err);
+        return NextResponse.json({ error: 'Failed to update config' }, { status: 500 });
+    }
+}
```
rahulkarajgikar: @coderabbitai do we need this file? i thought we are already setting the proxy in `next.config.ts`, could you explain to me what is the purpose of this file?

src/app/webui/components/ChatApp.tsx
```diff
@@ -0,0 +1,128 @@
...
+              <div className="space-y-2">
+                <Label htmlFor="exportName">Filename</Label>
+                <Input id="exportName" value={exportName} onChange={(e) => setExportName(e.target.value)} />
+              </div>
+              <DialogFooter>
+                <Button onClick={handleDownload}>Download YAML</Button>
+                <DialogClose asChild>
+                  <Button variant="ghost">Cancel</Button>
+                </DialogClose>
+              </DialogFooter>
```
rahulkarajgikar: maybe copy to clipboard is better than download for UX? then they can copy paste it into a separate file instead of having to check the downloaded yaml. what do you think? can update this separately

could even give both copy and download options actually

src/app/webui/eslint.config.mjs
rahulkarajgikar: actually let's do this in a separate PR

src/app/webui/components/ServersPanel.tsx
```diff
@@ -0,0 +1,319 @@
...
+  // other relevant details
+}
+// --- End of Interfaces ---
+
+interface ServersPanelProps {
+  isOpen: boolean;
+  onClose: () => void;
+  onOpenConnectModal: () => void;
+  variant?: 'overlay' | 'inline';
+}
```
rahulkarajgikar: let's not use duplicate types unless there is a good reason to - store types in `types.ts` or some common place and reuse them everywhere

src/app/webui/next.config.ts
rahulkarajgikar: can standardize all env variables for ports/server URLs separately in a different PR

src/app/webui/components/ServersPanel.tsx
rahulkarajgikar: if the types already exist in core, use those. if they don't exist, create re-usable types in `types.ts`

rahulkarajgikar (approved):
mostly LGTM, just a few minor things we should close:

- [can do in separate PR] standardizing express server/next js server URLs/ports with env variables
- [can do in separate PR] copy to clipboard option for config files
- [can do in separate PR] standardizing lint stuff - check coderabbit comment thread
- [can do in separate PR] potentially get rid of duplicate config/route api routes
- [do before merging] fixing duplicate types - reuse types from core if they exist. if they don't, create types in a reusable place like types.ts
- [do before merging] other minor comments i added on PR/renaming etc.

we can do the other stuff in separate PR to limit the scope of this pr but lets do it

src/app/api/server.ts
shaunak99: this updates the config when a new server is connected

src/app/api/server.ts
shaunak99: this exports the config with the updated server as shown above (have tested). I think it might be better to have a single source of truth for the config rather than maintain 2 different ones. Can think about this more

src/app/api/webui.ts
shaunak99: This only starts the express/api server, moved the function src/api/server.ts and renamed to startApiServer()

src/app/webui/eslint.config.mjs
shaunak99: Removed eslint in webui and made the update to follow base rules from root

src/app/webui/app/api/config/route.ts
shaunak99: removed

src/app/webui/components/ChatApp.tsx
shaunak99: Have added both options

PR #125, opened 2025-05-26
Enhance WebUI for better UX

shaunak99:
We have enhanced our webui for a cleaner experience. We have also added multiple features such as an MCP registry, model switching and the option to add/delete servers for better agent creation. We have also added a new 'mcp' mode in the startup command that enables starting a saiki instance explicitly as an mcp server

## Summary by CodeRabbit

- **New Features**
  - Added server registry browsing, searching, filtering, and installation via a new modal.
  - Enabled server deletion directly from the UI.
  - Introduced an LLM selector component with provider/model/router switching, API key input, and advanced options.
  - Added inline LLM selector with loading state and model switching in chat input.
  - Added keyboard shortcuts and a welcome screen with quick actions in chat interface.
  - Introduced a modal for adding custom server configurations.
  - Added execution history, quick input templates, and copy/share features in Playground.
  - Added a new example configuration for Notion workspace management.
  - Added support for a new "mcp" mode for agent operation.

- **Improvements**
  - Redesigned chat and server management UI with improved layout and controls.
  - Enhanced API with endpoints for resetting conversations, managing MCP servers, and querying/switching LLM configs.
  - Refined web UI startup logic for clearer production and development modes.
  - Improved logging for MCP server endpoints and tool interactions.
  - Updated styles, colors, and typography for a modern appearance.
  - Expanded documentation with a focused README for the web UI.
  - Updated package and build workflows with explicit web UI build and copy steps.

- **Bug Fixes**
  - Improved validation and error handling in LLM switching and configuration.
  - Fixed scroll behavior in message list to reliably scroll to bottom.

- **Chores**
  - Refactored build scripts and packaging workflow for the web UI.
  - Added new utility and UI components: badges, dropdown menus, tooltips, separators, scroll areas, and key-value editors.
  - Expanded and updated type definitions for chat, LLM, and server registry management.
  - Added internal API and client manager methods to support client removal and LLM switching.
  - Extended CLI option validation and tool confirmation provider for "mcp" mode.
  - Introduced a singleton service for server registry management with persistence and filtering.


src/app/api/server.ts
```diff
@@ -274,12 +334,154 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+                        case 'openai':
+                            return 'OpenAI';
+                        case 'anthropic':
+                            return 'Anthropic';
+                        case 'google':
+                            return 'Google';
+                        case 'groq':
+                            return 'Groq';
+                        default:
+                            return providerKey.charAt(0).toUpperCase() + providerKey.slice(1);
```
rahulkarajgikar: no need for a switch block with each provider, just one line with `providerKey.charAt(0).toUpperCase() + providerKey.slice(1);` is fine imo, that is anyway what we are doing

src/app/api/server.ts
```diff
@@ -274,12 +334,154 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+                const supportedRouters = (() => {
+                    switch (providerKey) {
+                        case 'openai':
+                        case 'anthropic':
+                            return ['vercel', 'in-built']; // Both routers supported
+                        case 'google':
+                        case 'groq':
+                            return ['vercel']; // Only vercel router supported
+                        default:
+                            return ['vercel']; // Default to vercel for unknown providers
```
rahulkarajgikar: we should move this metadata into the llm registry in core instead of here

that way we maintain one place for all information about routers, providers, llm, etc.

one function `getSupportedRoutersForProvider` could handle this

src/app/api/server.ts
```diff
@@ -274,12 +334,154 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+                    }
+                })();
+
+                // Only OpenAI supports custom baseURL for OpenAI-compatible endpoints
+                const supportsBaseURL = providerKey === 'openai';
```
rahulkarajgikar: same for this, it can be an entry in the llm registry map

and we can have a similar function `supportsBaseUrl(provider)` in the registry that reads the map and checks

src/app/webui/components/InputArea.tsx
```diff
@@ -3,18 +3,62 @@
...
   const textareaRef = useRef<HTMLTextAreaElement>(null);
   const [imageData, setImageData] = useState<{ base64: string; mimeType: string } | null>(null);
   const fileInputRef = useRef<HTMLInputElement>(null);
+  
+  // LLM selector state
+  const [currentModel, setCurrentModel] = useState('Loading...');
+  const [isLoadingModel, setIsLoadingModel] = useState(false);
+  
+  const coreModels = [
+    { name: 'Claude 4 Sonnet', provider: 'anthropic', model: 'claude-4-opus-20250514' },
```
rahulkarajgikar: minor issue: sonnet maps to opus

src/app/api/server.ts
```diff
@@ -274,12 +334,154 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+            }
+
+            // Validate provider/router combination
+            const providerInfo = LLM_REGISTRY[provider.toLowerCase()];
+            if (!providerInfo) {
+                return res.status(400).json({ error: `Unknown provider: ${provider}` });
+            }
+
+            const selectedRouter = router || 'vercel'; // Default to vercel if not specified
+
```
rahulkarajgikar: `validate provider router combination`  - one function in core - something like `isRouterSupportedForModel` can handle this

makes more sense to decide what is valid in the core than in the api layer

src/app/api/server.ts
```diff
@@ -274,12 +334,154 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+                    error: `Provider '${provider}' does not support 'in-built' router. Only OpenAI and Anthropic support in-built routing.`,
+                });
+            }
+
+            // Validate baseURL usage
+            if (baseURL && provider.toLowerCase() !== 'openai') {
+                return res.status(400).json({
+                    error: 'Custom baseURL is only supported for OpenAI provider',
+                });
+            }
```
rahulkarajgikar: similar here

src/core/ai/agent/SaikiAgent.ts
```diff
@@ -124,6 +126,67 @@ export class SaikiAgent {
...
     }
 
+    /**
+     * Gets the current LLM configuration and status.
+     * @returns Current LLM service configuration.
+     */
+    public getCurrentLLMConfig() {
+        return {
+            config: this.configManager.getConfig().llm,
+            serviceInfo: this.llmService.getConfig(),
```
rahulkarajgikar: why do we need this `serviceInfo`, remove if not used

src/core/ai/agent/SaikiAgent.ts
```diff
@@ -124,6 +126,67 @@ export class SaikiAgent {
...
+                this.clientManager,
+                this.agentEventBus,
+                this.messageManager // This preserves the conversation history
+            );
+
+            // Replace the LLM service using the private setter
+            this.setLLMService(newLLMService);
+
+            // Update the agent's config
+            this.configManager.getConfig().llm = newLLMConfig;
```
rahulkarajgikar: we shouldn't directly be setting things in the configManager over here, it's risky and hard to track updates this way

add helper function like this.configManager.updateLLMConfig() to do this

similar for any other places where we are directly mutating the config object fetched from configManager

src/core/ai/agent/SaikiAgent.ts
```diff
@@ -124,6 +126,67 @@ export class SaikiAgent {
...
+    private setLLMService(newLLMService: ILLMService): void {
+        // Use object assignment to update the readonly property safely
+        Object.defineProperty(this, 'llmService', {
+            value: newLLMService,
+            writable: false,
+            enumerable: true,
+            configurable: false,
+        });
+    }
+
```
rahulkarajgikar: `Object.defineProperty` is a hack to get around the variable being readonly.

instead, we should

1. remove `readonly` from the llmservice and make it `private` and prefix it with underscore.
```private _llmService: ILLMService```
3. change this setter function to a simple setter: `this.setLLMService(llmservice)` which does this.llmService = llmservice
4. add a getter function 
```
public getLLMService(): ILLMService {
    return this._llmService
}
```

File: src/core/config/manager.ts [hunk 10]
```diff
@@ -117,6 +178,16 @@ export class ConfigManager {
...
 
+    /**
+     * Type guard to check if a key is trackable in LLM provenance.
+     * @param key The key to check.
+     * @returns True if the key is a valid LLMOverrideKey that can be tracked in provenance.
+     */
+    private isTrackableKey(key: string): key is LLMOverrideKey {
+        const trackableKeys: LLMOverrideKey[] = ['provider', 'model', 'router', 'apiKey'];
+        return trackableKeys.includes(key as LLMOverrideKey);
+    }
```
rahulkarajgikar: @coderabbitai is this function necessary? would it be possible to just use LLMOverrideKey directly somehow, so that we just have one place to update instead of two? what do you think

src/app/api/server.ts
```diff
@@ -72,6 +78,23 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
         }
     });
 
+    // Reset to default configuration
+    app.post('/api/configs/reset', express.json(), async (req, res) => {
```
rahulkarajgikar: nit: api name could be better - /api/configs/reset implies some kind of config is getting reset but actually this API is just resetting the conversation

maybe /api/conversation/reset could be better

src/app/webui/types.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Shared type definitions for WebUI components
+ * Shared type definitions for Saiki - MCP tool testing playground
```
rahulkarajgikar, line 2: we are duplicating some types here which already exist in core

add a TODO to use the types from core instead

src/core/config/manager.ts
rahulkarajgikar: yeah we can do this, removing this function

rahulkarajgikar (approved):
few minor comments to address, rest lgtm

src/app/api/server.ts
shaunak99: duplicate endpoint, removing it

PR #134, opened 2025-05-31
Fix for multi-server, same name tools

shaunak99:
Currently, if you connect multiple servers and if they have the same toolNames, this leads to a conflict and overwrites the latest toolName, leading to a conflict. This is particularly challenging when connecting to multiple Saiki agents that are running as mcp servers. 

This fix resolve that by adding a server name at the beginning when there are multiple tools with the same name.

### DIFF
--- a/src/core/client/manager.ts
+++ b/src/core/client/manager.ts
@@ -11,12 +11,13 @@ export class MCPClientManager {
     private clients: Map<string, IMCPClient> = new Map();
     private connectionErrors: { [key: string]: string } = {};
     private toolToClientMap: Map<string, IMCPClient> = new Map();
+    private serverToolsMap: Map<string, Map<string, IMCPClient>> = new Map();
+    private toolConflicts: Set<string> = new Set();
     private promptToClientMap: Map<string, IMCPClient> = new Map();
     private resourceToClientMap: Map<string, IMCPClient> = new Map();
     private confirmationProvider?: ToolConfirmationProvider;
 
     constructor(confirmationProvider?: ToolConfirmationProvider) {
-        // If a confirmation provider is passed, use it, otherwise use the default implementation
         this.confirmationProvider = confirmationProvider ?? new CLIConfirmationProvider();
     }
 
@@ -40,6 +41,8 @@ export class MCPClientManager {
         const client = this.clients.get(clientName);
         if (!client) return;
 
+        this.serverToolsMap.delete(clientName);
+
         [this.toolToClientMap, this.promptToClientMap, this.resourceToClientMap].forEach(
             (cacheMap) => {
                 for (const [key, mappedClient] of cacheMap.entries()) {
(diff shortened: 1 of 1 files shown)


src/core/client/manager.ts
```diff
@@ -49,74 +52,148 @@ export class MCPClientManager {
...
         }
 
-        // Cache prompts, if supported
+        // Cache prompts and resources (unchanged)
         try {
             const prompts = await client.listPrompts();
             prompts.forEach((promptName) => {
                 this.promptToClientMap.set(promptName, client);
             });
-            logger.debug(`Cached prompts for client: ${clientName}`);
```
rahulkarajgikar: keep, don't delete. this is a useful log

src/core/client/manager.ts
```diff
@@ -49,74 +52,148 @@ export class MCPClientManager {
...
             logger.debug(`Skipping prompts for client ${clientName}: ${error}`);
         }
 
-        // Cache resources, if supported
         try {
             const resources = await client.listResources();
             resources.forEach((resourceUri) => {
                 this.resourceToClientMap.set(resourceUri, client);
             });
-            logger.debug(`Cached resources for client: ${clientName}`);
```
rahulkarajgikar: keep, don't delete. this is a useful log

src/core/client/manager.ts
```diff
@@ -49,74 +52,148 @@ export class MCPClientManager {
...
+        // Update main tool map - remove conflicted tools
+        for (const conflictedTool of this.toolConflicts) {
+            this.toolToClientMap.delete(conflictedTool);
+        }
+    }
+
     private async updateClientCache(clientName: string, client: IMCPClient): Promise<void> {
-        // Cache tools
+        const serverTools = new Map<string, IMCPClient>();
+        this.serverToolsMap.set(clientName, serverTools);
```
rahulkarajgikar, line 160: @coderabbitai what is the point of these 2 lines? i'm not understanding what this.serverTooldMap is doing here - why is the key just a map. can you explain how this works?

src/core/client/manager.ts
```diff
@@ -49,74 +52,148 @@ export class MCPClientManager {
...
+            const actualToolName = toolName.substring(firstUnderscoreIndex + 1);
+
+            // Find server by matching prefix
+            for (const [serverName, serverTools] of this.serverToolsMap.entries()) {
+                const normalizedServerName = serverName.replace(/[^a-zA-Z0-9]/g, '_');
+                if (normalizedServerName === serverPrefix && serverTools.has(actualToolName)) {
+                    return serverTools.get(actualToolName);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
rahulkarajgikar: agree, we should have a better way of figuring out if a tool is server prefixed or not
this is very brittle and 90%+ of tools will have underscores

what do you suggest? since this goes directly into the LLM we also don't want to add too much fluff in the function name, let's keep that in mind.

we should have a function for checking if a tool name has been "server prefixed" and rely on that to provide this check and/or a function that returns the actual tool name given any name.


i think some better primitives would help here

src/core/client/manager.ts
```diff
@@ -267,23 +332,20 @@ export class MCPClientManager {
...
      * Get all registered clients
-     * @returns Map of client names to client instances
      */
     getClients(): Map<string, IMCPClient> {
         return this.clients;
     }
 
     /**
      * Get the errors from failed connections
-     * @returns Map of server names to error messages
```
rahulkarajgikar: keep this

src/core/client/manager.ts
```diff
@@ -267,23 +332,20 @@ export class MCPClientManager {
 
     /**
      * Get all registered clients
-     * @returns Map of client names to client instances
```
rahulkarajgikar: keep this

src/core/client/manager.ts
```diff
@@ -267,23 +332,20 @@ export class MCPClientManager {
...
      * Get the errors from failed connections
-     * @returns Map of server names to error messages
      */
     getFailedConnections(): { [key: string]: string } {
         return this.connectionErrors;
     }
 
     /**
      * Disconnect and remove a specific client by name.
-     * @param name The name of the client to remove.
```
rahulkarajgikar: keep

src/core/client/manager.ts
```diff
@@ -49,74 +52,148 @@ export class MCPClientManager {
...
+                            ...toolDef,
+                            description: toolDef.description
+                                ? `${toolDef.description} (via ${serverName})`
+                                : `Tool from ${serverName}`,
+                        };
+                    }
+                }
+            }
+        }
+
```
rahulkarajgikar: would it make sense to inverse this logic?

essentially we know we only have to iterate through all the names in this.toolConflicts - these are the conflicted tools

so maybe it could reduce search complexity if we iterated first through all the names in toolConflicts?

@coderabbitai is this correct, or would the time complexity be the same either way?

src/core/client/manager.ts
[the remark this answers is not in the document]
rahulkarajgikar: makes sense! checking if a tool name is in the conflicts map is a reliable way of determining if it actually is server prefixed or not

src/core/client/manager.ts
rahulkarajgikar: cool, this makes sense. thanks!

src/core/client/manager.ts
```diff
@@ -49,74 +52,148 @@ export class MCPClientManager {
...
+            if (count > 1) {
+                this.toolConflicts.add(toolName);
+            }
+        }
+
+        // Update main tool map - remove conflicted tools
+        for (const conflictedTool of this.toolConflicts) {
+            this.toolToClientMap.delete(conflictedTool);
+        }
+    }
```
rahulkarajgikar: @coderabbitai could we optimize this? 

the problem here is that each time any client cache is cleared we rebuild the entire tool conflicts each time.
having a rebuild method is fine, but for the case of clearClientCache - technically we only need to remove conflicting tool if a client is removed.

so ideally, we would first check which client is removed, see if any tool of that server/client is a reason for conflict, and simply remove that. although an edge case could be if 3 or more servers all had the same tool name - this wouldn't work but a rebuild is more robust.

now that i think of this, it seems like rebuilding should be fine, we mostly wont have more than 100 tools anyway

what do you think?

rahulkarajgikar: alright let's keep this the way it is

src/core/client/manager.ts
rahulkarajgikar: got it

src/core/client/manager.ts
rahulkarajgikar: yes, getAllTools() is a hot path used on every LLM invocation so performance would matter

shaunak99:
@coderabbitai review

shaunak99:
@rahulkarajgikar need this to be fixed for enabling multi-agent systems like in #219 or else `chat_with_agent` has conflicts

shaunak99:
True but the question comes do you change this nomenclature internally or upon connection? 

This would only solve this for internal use cases where saiki agents are being used. But for a situation where you have two servers, say `server_1` & `server_2` that both expose `search()` as a function, you need to be able to map the `search()` function to the right server when executing. In our current code, this would have always executed the first function.

So you still need some delimiter based on logic or equivalent solution like this to decompose and call the relevant function. The LLM would see it as something like `server_1_search()` and `server_2_search()` as it does even now. Choice of delimited should also be such that it reduces chances of failed parsing as initially suggested by coderabbit. 

Ideally, we should have generic logic to handle this across server instead of fitting it for saiki specific agents. A better solution to this overall can definitely be explored, but is needed for sure.

PR #153, opened 2025-06-16
updated storage backend

shaunak99:
Allows each agent instance to have an independent storage backend

## Summary by CodeRabbit

- **Documentation**
  - Updated usage examples to reflect the new storage initialization API.

- **Refactor**
  - Simplified storage management by removing deprecated singleton functions.
  - Introduced a new method to create multiple independent storage manager instances.

- **Chores**
  - Updated internal references to use the new storage initialization method.

### DIFF
--- a/src/core/storage/storage-manager.ts
+++ b/src/core/storage/storage-manager.ts
@@ -218,21 +218,6 @@ export class StorageManager {
     }
 }
 
-// Default storage manager instance
-let defaultStorageManager: StorageManager | null = null;
-
-/**
- * Initialize the default storage manager with configuration
- */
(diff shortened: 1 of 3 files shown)


src/core/storage/storage-manager.ts
```diff
@@ -233,6 +233,15 @@ export async function initializeStorage(config: StorageConfig): Promise<StorageB
     return await defaultStorageManager.connect();
```
rahulkarajgikar: nit: we can also get rid of `defaultStorageManager` and the functions that are using it completely

rahulkarajgikar:
just check coderabbit nits once, it has a good point about returning back the manager

PR #170, opened 2025-06-19
Rename --config-file to --agent

rahulkarajgikar:
also rename `saiki.yml` to `agent.yml`

and `configuration` folder to `agents`


rahulkarajgikar:
@shaunak99 i updated `saiki.yml` to `agent.yml` and changed the agent instructions to make it Jarvis instead of Saiki.

lmk if you think this is fine 

i think we could also consider renaming `agent.yml` to `default.yml` or even `jarvis.yml` or something else

also yeah we need to make changes in the web ui as well (to not say Hi, i am saiki, talk to me type language)

shaunak99:
I think it's better to have the default one without any name if not Saiki or else it suffers from the same problem.

Something neutral like `agent.yml`, `assistant.yml`, `default.yml`.

The default persona in the system prompt also becomes "You are an AI assistant" and I think this is also the default response you get when you directly call the LLM without any system prompting.

rahulkarajgikar:
that makes sense, i think i'll make it `agent.yml` for now and just get rid of the name Jarvis, can revisit later

rahulkarajgikar:
rebased and updated new docs as well

rahulkarajgikar:
@coderabbitai review

PR #175, opened 2025-06-20
Refactor SaikiAgent usage to be more siimlar to other libraries

rahulkarajgikar:
This changes the preferred method of using the agents via library

old:

```
const agent = await createSaikiAgent()...
const response = await agent.run()
```

new:
```
const agent = new SaikiAgent()
await agent.start()


const response = await agent.run()

// once done:
await agent.stop()
```

updates all documentation, cli scaffolding commands, and adds new tests


rahulkarajgikar:
moving to #176, closing this

PR #211, opened 2025-07-05
Handle ollama failures for hosted gemma3

rahulkarajgikar:
hosted gemma3n:e2b on ollama doesn't support tools 

Updated generateText and streamText methods in vercel router to make a small test call to see if the model supports tools (checking for exact error string) before processing the request.

exact error string check is a bit brittle but there's no cleaner way

This information is cached so we don't need to repeat it after the first response

tested webui and cli streaming/non streaming

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility by automatically detecting whether the current AI model supports tools, preventing errors when tools are not supported.

* **Refactor**
  * Enhanced performance with caching for tool support checks, reducing unnecessary repeated validations.

### DIFF
--- a/src/core/ai/llm/services/vercel.ts
+++ b/src/core/ai/llm/services/vercel.ts
@@ -103,7 +103,9 @@ export class VercelLLMService implements ILLMService {
             `[VercelLLMService] Tools before formatting: ${JSON.stringify(tools, null, 2)}`
         );
 
+        // If the model supports tools, format them otherwise use an empty object
         const formattedTools = this.formatTools(tools);
+
         logger.silly(
             `[VercelLLMService] Formatted tools: ${JSON.stringify(formattedTools, null, 2)}`
         );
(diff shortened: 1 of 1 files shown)


shaunak99:
This feels like a patch for an edge case, let's not do it this way. If there's a way to identify the model (like vercel and openai sdk I think throws an error for this), let's use that or else let's have this check only for non-registered models since that's the only case we would not know.

rahulkarajgikar:
agree, but there's not really a better way atm

vercel is just passing on the error that ollama is giving at the custom base URL, there's no custom exception we can use.

vercel ai sdk doesn't have a way to identify if the model supports tools

we only have the error string to work with - we have to handle this case-by-case

even for non-registered models, the errors could vary based on how they are hosted so we can't really generalize

if we see more errors like this, it's a quick fix until we have a better solution

not really worth indexing on a generalized solution either until we have enough reason to do so

PR #213, opened 2025-07-06
remove reset conversation in headless mode

rahulkarajgikar:
this allows headless CLI to remember previous interactions


rahulkarajgikar:
example:

saiki "give me a command to kill process running on http://localhost:11434/"
<saiki responds back with question>
saiki "yes, do this"

PR #237, opened 2025-07-22
bug: mask apiKey in API responses to enhance security

### LINKED ISSUE #236: Bug: API is returning apikey in public response
**Describe the bug**  
The model API is returning an `apikey` field in its response payload. This is a security concern, as API keys should be kept confidential and not exposed to clients.

---

**Expected behavior**  
The `apikey` field should not be included in any public API response. It should be kept server-side only or marked as `write_only` in the serializer to prevent leaking it through the API.

Yadavanurag13:
Closes #236 

## Summary by CodeRabbit

* **Bug Fixes**
  * API responses for LLM configuration now mask the API key field to prevent exposure of sensitive information.

### DIFF
--- a/src/app/api/server.ts
+++ b/src/app/api/server.ts
@@ -449,6 +449,11 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
                 ? agent.getEffectiveConfig(sessionId as string).llm
                 : agent.getCurrentLLMConfig();
 
+            // Mask apiKey before returning
+            if (currentConfig && currentConfig.apiKey) {
+                currentConfig.apiKey = '[REDACTED]';
+            }
+
             res.json({ config: currentConfig });
         } catch (error: any) {
             logger.error(`Error getting current LLM config: ${error.message}`);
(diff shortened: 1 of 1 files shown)


src/app/api/server.ts
```diff
@@ -508,6 +513,10 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
             Object.assign(llmConfig, otherFields);
 
             const result = await agent.switchLLM(llmConfig, sessionId);
+            // Mask apiKey in result if present
+            if (result && result.config && result.config.apiKey) {
```
rahulkarajgikar: can use result?.config?.apiKey to avoid &&, similar in above if statement

rahulkarajgikar (changes requested):
minor changes requested

rahulkarajgikar:
long term better solution to protect all APIs using a middleware wrapper:

we have a helper `sendJsonResponse` that we use to support pretty printing that is used by few of the APIs

rough plan

1. enhance this to make it `sendSecureJsonResponse` - which also handles redacting/removing sensitive info from the response before sending
2. migrate all APIs which aren't using this to all use this 
3. test APIs with curl and web UI to see that they are working properly

@Yadavanurag13 let me know if you want to explore this and we can make a separate issue for this and brainstorm some ideas

yanurag-dev:
@rahulkarajgikar I would like to explore. Just wanted to know should we quick fix it for now.

rahulkarajgikar:
yeah we will merge this PR to quick fix for now.
we can explore a long term solution separately in a separate issue + follow up PR

src/app/api/server.ts
yanurag-dev: I have pushed the quick fix. review it

yanurag-dev:
@rahulkarajgikar Pushed the requested changes.

PR #239, opened 2025-07-22
Custom Tools

shaunak99:
Introduces a way to define and add custom tools to agent alongside MCPServers.

## Summary by CodeRabbit

* **New Features**
  * Introduced a unified tool management system, allowing discovery and execution of both MCP and internal tools via a new ToolManager.
  * Added support for internal tools, including a new "search_history" tool for searching conversation history.
  * CLI and API updated to display and handle both MCP and internal tools.
  * SaikiAgent now provides unified methods for tool execution and discovery from all sources.

* **Bug Fixes**
  * Improved error messages and documentation for tool execution and management.

* **Documentation**
  * Updated API docs to reflect unified tool management and new methods.

* **Tests**
  * Added comprehensive tests for ToolManager, internal tools configuration, and integration scenarios.

* **Chores**
  * Refactored codebase to use ToolManager instead of direct MCPManager references.


src/core/tools/types.ts
```diff
@@ -0,0 +1,141 @@
+import { z } from 'zod';
```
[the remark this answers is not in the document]
shaunak99: The zod import is already correct - it's an external npm package, not a local file, so it doesn't need the .js extension. The coding guidelines about .js extensions only apply to local file imports, not npm packages.

tools/math.ts
```diff
@@ -0,0 +1,73 @@
+import { createTool } from '@truffle-ai/saiki';
+import { z } from 'zod';
```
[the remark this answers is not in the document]
shaunak99: Current import is intended for example purposes with the published package.

src/app/cli/commands.ts
```diff
@@ -835,11 +835,22 @@ export const CLI_COMMANDS: CommandDefinition[] = [
 
                 // Tools
                 try {
-                    const tools = await agent.mcpManager.getAllTools();
+                    const allTools = await agent.getAllTools();
+                    const mcpTools = await agent.getAllMcpTools();
+                    const customTools = Object.keys(allTools).filter(
```
rahulkarajgikar: this can be a separate func `agent.getCustomTools()`

this would simplify filtering logic which we are repeating both here and below in /tools cmd and in cli.ts

src/app/cli/cli.ts
```diff
@@ -73,11 +73,14 @@ async function _initCli(agent: SaikiAgent): Promise<void> {
     // Load available tools
     logger.info('Loading available tools...');
     try {
-        const tools = await agent.mcpManager.getAllTools(); // tools variable is not used currently but kept for potential future use
+        const allTools = await agent.getAllTools();
+        const mcpTools = await agent.getAllMcpTools();
+        const customTools = Object.keys(allTools).filter(
+            (name) => !Object.keys(mcpTools).includes(name)
+        );
```
rahulkarajgikar: we can make a custom tools func (check below)

src/core/ai/agent/SaikiAgent.ts
```diff
@@ -914,8 +914,8 @@ export class SaikiAgent {
                 success: true,
             });
             this.agentEventBus.emit('saiki:availableToolsUpdated', {
-                tools: Object.keys(await this.mcpManager.getAllTools()),
-                source: 'mcp',
+                tools: Object.keys(await this.services.toolManager.getAllTools()),
+                source: 'unified',
```
rahulkarajgikar: source would still be mcp, if we are connecting an mcp server

src/core/ai/agent/SaikiAgent.ts
```diff
@@ -959,6 +959,18 @@ export class SaikiAgent {
...
     }
 
+    /**
+     * Executes a tool from any source (MCP servers or custom tools).
+     * This is the unified interface for tool execution that can handle both MCP and custom tools.
+     * @param toolName The name of the tool to execute
+     * @param args The arguments to pass to the tool
+     * @returns The result of the tool execution
+     */
+    public async executeTool(toolName: string, args: any): Promise<any> {
```
rahulkarajgikar, line 961: likely dupe of `executeMcpTool` - we can remove `executeMcpTool` above and just keep this new func

src/core/ai/agent/SaikiAgent.ts
```diff
@@ -959,6 +959,18 @@ export class SaikiAgent {
...
+        this.ensureStarted();
+        return await this.services.toolManager.executeTool(toolName, args);
```
rahulkarajgikar: make it `this.toolManager` - not sure why we are going through services, not consistent with other patterns

src/core/config/schemas.ts
```diff
@@ -587,6 +587,62 @@ export const StorageSchema = z
...
+        enabledTools: z
+            .union([z.literal('all'), z.array(z.string())])
+            .default('all')
+            .describe('Enable all tools with "all" or specify array of tool IDs to enable'),
+
+        // Tool discovery options
+        enableToolDiscovery: z
+            .boolean()
+            .default(true)
+            .describe('Whether to automatically discover tools from the tools/ directory'),
```
rahulkarajgikar: i think this should be enforced, always true, not optional

src/core/config/schemas.ts
```diff
@@ -587,6 +587,62 @@ export const StorageSchema = z
...
+
+        // Tool-specific configurations
+        toolConfigs: z
+            .record(z.record(z.any()))
```
rahulkarajgikar: we need type for this, can't use any. unintuitive about what can be configured

src/core/config/schemas.ts
```diff
@@ -587,6 +587,62 @@ export const StorageSchema = z
...
+                timeout: z
+                    .number()
+                    .int()
+                    .positive()
+                    .optional()
+                    .describe('Default execution timeout for custom tools in milliseconds'),
+                enableCaching: z
+                    .boolean()
+                    .default(false)
+                    .describe('Whether to enable result caching for custom tools'),
```
rahulkarajgikar: get rid of this, we aren't using it anywhere and don't need it

src/core/config/schemas.ts
```diff
@@ -587,6 +587,62 @@ export const StorageSchema = z
 
 export type StorageConfig = z.infer<typeof StorageSchema>;
 
+export const CustomToolsConfigSchema = z
+    .object({
+        // Tool filtering options
+        enabledTools: z
+            .union([z.literal('all'), z.array(z.string())])
+            .default('all')
```
rahulkarajgikar: not a fan of `all` -> explicit names are better and easier to understand and debug

the issue is we need some kind of validation/check that the tool is a valid name

src/core/tools/custom-tools-provider.test.ts
```diff
@@ -0,0 +1,327 @@
...
+                {
+                    enabledTools: 'all',
+                    enableToolDiscovery: false, // Disable tool discovery for testing
+                    globalSettings: { enableCaching: false },
+                },
+                confirmationProvider
+            );
+
+            expect(provider).toBeDefined();
+            expect(provider.getToolNames()).toEqual([]);
```
rahulkarajgikar: shouldn't this be pulling all the tools?

src/core/tools/custom-tools-provider.test.ts
```diff
@@ -0,0 +1,327 @@
...
+                globalSettings: {
+                    requiresConfirmation: true,
+                    timeout: 5000,
+                    enableCaching: false,
+                },
+            };
+
+            provider = new CustomToolsProvider(config, confirmationProvider);
+
+            expect(provider).toBeDefined();
```
rahulkarajgikar: we should be expecting provider.getToolNames == above tools also

src/core/tools/custom-tools-provider.ts
```diff
@@ -0,0 +1,204 @@
...
+import { ToolConfirmationProvider } from '../client/tool-confirmation/types.js';
+import { logger } from '../logger/index.js';
+import { join } from 'path';
+
+/**
+ * Provides custom tool functionality using dedicated service classes
+ *
+ * This provider follows the codebase pattern of being a thin orchestrator
+ * that delegates to specialized service classes for custom tools
+ */
```
rahulkarajgikar: we should describe what the class actually does

src/core/tools/tool-executor.ts
```diff
@@ -0,0 +1,262 @@
...
+    toolId: string;
+    args: ToolExecutionArgs;
+    context?: ToolExecutionContext;
+}
+
+/**
+ * Tool execution response interface
+ * Represents the standardized response format for tool execution
+ */
+export interface ToolExecutionResponse {
```
rahulkarajgikar: this looks the same as ToolExecutionResult, keep just 1

src/core/tools/tool-executor.ts
```diff
@@ -0,0 +1,262 @@
...
+    success: boolean;
+    data?: unknown;
+    error?: string;
+    metadata?: Record<string, unknown> | undefined;
+}
+
+/**
+ * No-op confirmation provider for testing
+ */
+class NoOpConfirmationProvider implements ToolConfirmationProvider {
```
rahulkarajgikar: we already have something like this, if not, move to tool confirmation provider folder

src/core/tools/types.ts
```diff
@@ -0,0 +1,134 @@
...
+    id: string;
+
+    /** Human-readable description of what the tool does */
+    description: string;
+
+    /** Zod schema defining the input parameters */
+    inputSchema: z.ZodSchema;
+
+    /** The actual function that executes the tool */
+    execute: (input: any, context?: ToolExecutionContext) => Promise<any> | any;
```
rahulkarajgikar, line 23: input can be any but response should be ToolExecutionResponse or equivalent

src/core/utils/service-initializer.ts
```diff
@@ -94,30 +96,38 @@ export async function createAgentServices(
...
     logger.debug('Agent state manager initialized');
 
-    // 7. Initialize session manager
+    // 8. Initialize session manager
     const sessionManager = new SessionManager(
         {
             stateManager,
             promptManager,
-            mcpManager,
+            toolManager,
```
rahulkarajgikar, line 148: with this change, we move `mcpManager` inside `toolManager`

it is probably better to keep them standalone, while still giving `toolManager` access to the `mcpManager` when initializing it so that it has access to the mcp tools. 

the reason is that MCPManager handles a lot more than just tools (servers + prompts/resources/etc. for each server). so having `mcpManager` and `toolManager` both as first class services just fits the model better

so for anything tools related -> go through agent.toolManager
for anything MCP related -> go through agent.mcpManager

there might be some overlap -> mcpManager.getAllTools() will get all mcp tools
and we wouuld also have an equivalent toolsManager.getMcpTools() that just calls the mcpManager.getAllTools()

this level of duplication is ok and worth the separation of concerns -> this way we have 2 services at the same level (MCPManager, ToolsManager, and ToolsManager has a dependency on MCPManager)

src/core/tools/custom-tools-provider.test.ts
```diff
@@ -0,0 +1,327 @@
...
+                confirmationProvider
+            );
+
+            await provider.initialize();
+
+            const toolNames = provider.getToolNames();
+            expect(toolNames).toHaveLength(0);
+        });
+
+        it('should handle non-existent tool IDs in enabledTools', async () => {
```
rahulkarajgikar: this should fail early with some exception instead of silently skipping

src/core/tools/tool-execution.e2e.test.ts
```diff
@@ -0,0 +1,561 @@
...
+import { ToolManager } from './tool-manager.js';
+import { NoOpConfirmationProvider } from '../client/tool-confirmation/noop-confirmation-provider.js';
+import type { ToolSet } from './types.js';
+
+/**
+ * End-to-end tests for tool execution flows
+ * These tests simulate real-world tool usage scenarios
+ */
+
+// Mock MCP Manager for E2E tests
```
rahulkarajgikar: use `.integration.test.ts` format, not `.e2e.test.ts`

src/core/tools/tool-execution.e2e.test.ts
```diff
@@ -0,0 +1,561 @@
...
+    let toolManager: ToolManager;
+    let customToolProvider: any;
+    let mcpManager: E2EMCPManager;
+    let confirmationProvider: NoOpConfirmationProvider;
+
+    beforeEach(async () => {
+        confirmationProvider = new NoOpConfirmationProvider();
+
+        // Mock the customToolProvider with simpler implementation
+        customToolProvider = {
```
rahulkarajgikar: @coderabbitai can you explain if this should be mocked or not? i am not understanding, imo i feel like this should use real code, can you explain

src/core/tools/tool-manager.ts
```diff
@@ -0,0 +1,419 @@
...
+            logger.info(
+                `Custom tools initialized: ${this.customToolProvider.getToolNames().length} tools available`
+            );
+        }
+    }
+
+    /**
+     * Normalize tool parameters to ensure consistent structure
+     */
+    private normalizeToolParameters(toolDef: any): any {
```
rahulkarajgikar: avoid `any`

src/core/tools/schema-converter.ts
```diff
@@ -0,0 +1,237 @@
...
+    additionalProperties?: JSONSchema | boolean;
+    anyOf?: JSONSchema[];
+}
+
+/**
+ * Converts Zod schemas to JSON Schema format for MCP compatibility
+ *
+ * Handles common Zod types used in tool definitions
+ */
+export class SchemaConverter {
```
rahulkarajgikar: @claude do we already have something like this in the codebase

rahulkarajgikar (commented):
have reviewed about 60-70%

a lot of this current design in the core backend is good, and the idea of enforcing the tools in a specific location is good, avoiding relative paths (extremely painful)

but my hunch is that dynamically loading modules from tools/directory at runtime will be quite error prone and unreliable - i had similar problems while working on plugins PR, and the whole design ended up being super brittle and unreliable. some of the problems:
 - errors in code for custom tools won't be caught at compile time, because we never actually compile the code for the tool
 - incorrect tool names in YML also won't be caught early
 - this will lead to having to actually load up the agent each time, and testing out a specific prompt for the model that uses that tool, for something which could be caught at compile time. there are some ways to catch some of this earlier, but the primary issue here is that ts code should be getting compiled to js and validated as part of a step before the agent runs at all
 - dependence on current working directory to resolve tools is brittle and ideally shouldn't be there - it enforces users to run commands only from a specific location which can easily be missed by users (they walk up a directory, re-run CLI with updated filepaths, everything now breaks)
 
to solve these problems, have been working on this design which aims to solve a lot of these issues: 
https://www.notion.so/Saiki-revamped-CLI-design-23e63b89f2dd80d8b60eea69aa252ff5?pvs=12 - 

it introduces a fixed project structure + an intermediate `build` step which is responsible for handling custom code (like custom tools). this also gives us a clear way to think about the CLI (inside a saiki project v/s outside the saiki project)

we enforce users to get started with saiki create-app, enforce locations with CLI scaffold commands, and catch custom code issues at compile time with saiki build, kind of like next-js

instead of having cwd based file detection, we use project based file detection to understand if the cwd is inside a saiki project (check current path.ts code -> we recursively walk up from cwd until we hit a package.json that has a dependency on @ truffle-ai/saiki.)

this moves away from having a global CLI that can run anywhere, but gives us a much more robust useful CLI, solves the custom code problem, and gives users a well-defined project structure for them to experiment with agents

i am planning to move towards this design, lmk your thoughts.

we can likely re-use a lot of the code in this PR, but some of the modules will be completely different (mostly path resolution/tool building logic)

shaunak99:
Appreciate the detailed feedback! Based on our discussion, I have removed the `customTools` system and added only the essential parts needed for internal-tools while preserving the `ToolManager` implementation. I have also tried to update the MCPManager and ToolManager to keep them as first-class services. Let me know if any changes need to be made here. 

Additionally, I have added an internal-tool using the `SearchService` to enable the agent to have the ability to search & reference past session interactions using keyword search while chatting. Will be adding a similar service for scheduler in future PR.

src/core/tools/types.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Tool manager's unified tool set (for compatibility with MCP tools and ai/types.ts)
+ */
+export interface ToolManagerToolSet {
+    [toolName: string]: {
+        name?: string;
+        description: string;
+        parameters?: ToolParameters;
+    };
+}
```
rahulkarajgikar, line 57: we already have most of these types in this file in core/ai/types.ts

we can move those types into here and add the new ones, these are mostly duplicate

there are also some small subtle changes in these duplicate types that are leading to a whole lot of unnecessary code - `ToolManagerToolSet` is the same as `ToolSet` except that description is no longer optional.

this is resulting in `ToolManager.getMcpTools()`, which should be just one line calling `McpManager.getAllTools()` needing unnecessary type conversion between the two ToolSet types. just using ToolSet everywhere would have avoided this

there are some other problems, i will fix in follow up

src/core/tools/types.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export interface Tool {
+    /** Unique identifier for the tool */
+    id: string;
+
+    /** Human-readable description of what the tool does */
+    description: string;
+
+    /** Zod schema defining the input parameters */
+    inputSchema: any; // Using any for flexibility since we don't import zod here
```
rahulkarajgikar, line 20: import zod here and don't use any

src/core/utils/service-initializer.ts
```diff
@@ -94,30 +99,53 @@ export async function createAgentServices(
...
+
+    // 6. Initialize internal tools provider with services and confirmation support
+    const internalToolsProvider = new InternalToolsProvider(
+        {
+            searchService,
+            // Future services can be added here as needed
+        },
+        confirmationProvider,
+        config.internalTools
+    );
```
rahulkarajgikar, line 113: can be internal dependency inside ToolManager

rahulkarajgikar (approved):
tested it out, works well!

there are some duplicate types and unnecessary code in the tools module which can be simplified, i'll merge this and fix it in a follow up

PR #242, opened 2025-07-23
Feature: Added support for File Parts to enable complete multimodal inputs

shaunak99:
This adds support for the LLM to support File Parts. Gemini and OpenAI (gpt-4o-audio-preview) support sending audio files. Gemini, OpenAI & Anthropic support sending PDFs natively (processed as images, page by page). 

Screenshot 2025-07-24 at 2 46 54 AM

Also updated WebUI to support the same.

## Summary by CodeRabbit

* **New Features**
  * Added support for sending and receiving file attachments (PDFs and audio files) in chat conversations.
  * Enabled audio recording directly within the chat input area.

* **Enhancements**
  * Chat messages now visually display and allow playback of audio files and show attached PDF files.
  * Updated chat input and message list UI to support and preview file attachments.
  * Expanded API documentation to detail supported file types and file upload capabilities.
  * Added model-aware filtering and validation for file attachments based on provider and model capabilities.
  * Improved message formatting to handle file content across multiple LLM providers.
  * Enhanced type safety and tooling around file data in chat and AI message handling.
  * Added comprehensive input validation for text, image, and file data with user feedback on unsupported inputs.
  * Introduced event signaling for unsupported input scenarios in LLM services.

* **Bug Fixes**
  * Adjusted tests to align with updated method signatures for file support.

* **Documentation**
  * Updated API and type documentation to reflect new file attachment features and supported formats.

### DIFF
--- a/src/core/ai/session/chat-session.ts
+++ b/src/core/ai/session/chat-session.ts
@@ -242,7 +242,7 @@ export class ChatSession {
             `Running session ${this.id} with input: ${input}, imageDataInput: ${imageDataInput}, fileDataInput: ${fileDataInput}`
         );
 
-        // Validate input for SDK users (API layer also validates, but SDK users bypass API)
+        // Validate input at session level
         const currentConfig = this.services.stateManager.getLLMConfig(this.id);
         const validation = validateInputForLLM(
             {


src/app/api/server.ts
```diff
@@ -84,7 +84,50 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+                // Security validation: base64 format check
+                if (typeof base64 === 'string') {
+                    const base64Regex = /^[A-Za-z0-9+/]*={0,2}$/;
+                    if (!base64Regex.test(base64)) {
+                        return res.status(400).send({ error: 'Invalid file data format' });
+                    }
+                }
+
+                fileDataInput = { data: base64, mimeType, filename };
+            }
```
rahulkarajgikar: we can move this to a helper function `extractFileInputFromRequest` and call it in all of these apis, code is repeated 2 times

more re-usable, less code

also we can move `allowedmimetypes` to a global constant in the file `ALLOWED_MIME_TYPES `so that it can be re-used in the websocket response and we have one definitive source of truth for the allowed input types

we will still need a duplicate impl for the websocket API though, which is fine

actually coderabbit has a similar comment with the suggestion for implementation in it's recent review, duplicate comments

src/app/api/server.ts
```diff
@@ -84,7 +84,50 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+
+                // Security validation: file size check (max 50MB for base64)
+                if (typeof base64 === 'string' && base64.length > 67108864) {
+                    // ~50MB in base64
+                    return res.status(400).send({ error: 'File size too large (max 50MB)' });
+                }
+
+                // Security validation: MIME type allowlist
+                const allowedMimeTypes = [
+                    'application/pdf',
```
rahulkarajgikar: could we also support text file formats? is that included in pdf itself or is that a different file extension

would be useful (i upload a lot of .txt, .md, .doc, .docx) files into chatgpt

src/app/webui/components/MessageList.tsx
```diff
@@ -197,13 +197,48 @@ export default function MessageList({ messages }: MessageListProps) {
                             />
                           );
                         }
+                        if ((part as any).type === 'file' && 'data' in part && 'mimeType' in part) {
+                          const filePart = part as any;
```
rahulkarajgikar, line 225: avoid any type

src/core/ai/agent/SaikiAgent.ts
```diff
@@ -274,13 +274,15 @@ export class SaikiAgent {
      *
      * @param userInput - The user's message or query to process
      * @param imageDataInput - Optional image data and MIME type for multimodal input
+     * @param fileDataInput - Optional file data and MIME type for file input
      * @param sessionId - Optional session ID for multi-session scenarios
      * @returns Promise that resolves to the AI's response text, or null if no significant response
      * @throws Error if processing fails
      */
     public async run(
         userInput: string,
```
rahulkarajgikar: might be good to rename `userInput` to `textInput` which is more accurate

src/core/ai/llm/services/vercel.ts
```diff
@@ -152,13 +152,14 @@ export class VercelLLMService implements ILLMService {
...
         stream?: boolean
     ): Promise<string> {
-        // Add user message, with optional image data
+        // Add user message, with optional image and file data
         logger.debug(
-            `VercelLLMService: Adding user message: ${userInput} and imageData: ${imageData}`
+            `VercelLLMService: Adding user message: ${userInput}, imageData: ${imageData}, fileData: ${fileData}`
         );
-        await this.contextManager.addUserMessage(userInput, imageData);
+        await this.contextManager.addUserMessage(userInput, imageData, fileData);
```
rahulkarajgikar: we need some kind of check either here, or earlier in the flow of control that checks if these input formats are  actually supported for the current LLM. if it isn't, we can't add it into the message history. 

ex: if we upload an audio file and the current LLM doesn't support it, currently we don't have a check, so the response generation will fail, and since the unsupported input is in the message history, it will render the chat unusable since the entire message history is sent into the LLM each time.

the best way i can think of is to add metadata about each LLM model into the LLM registry (supportsAudio, supportsPDF, or supportedFileTypes) and check that here (or even earlier is better). if we reach this point in the code and the llm doesn't support the data it's getting, we need to throw some exception, emit an event and handle the exception in calling code

best designs in general block incorrect inputs as early as possible, at the source (web UI or API). some thought dump around this:
- add supported input metadata into llm registry
- expose llm registry info via some API
- call this API in web UI (we currently have a hard coded list of models in web UI, with this web UI could now know all allowed models)
- based on the current LLM (which web UI would already know), only allow files that are supported by that LLM to be uploaded in the web UI. this prevents web UI users from accidentally uploading unsupported files at all, eliminating 99% of possible issues
- API routes responsible for generation can also pre-filter bad/unsupported inputs based on the current LLM and return 4xx error codes early on

for the API routes generating responses:
 - we are currently accepting imagedata/filedata as input, and passing it through the core without any checks
 - since we know the current llm for the agent when the API is hit, we can block bad inputs early at the API before it even hits the core by checking the metadata for the current LLM
 - maybe this logic could be added to the previous code block we have for checking supported file types in the API routes, we could have supported file types per LLM and based on LLM, allow that through? 
 - in case input is unsupported, return 400 error code in the API (unsupported input, ex: LLM <x> does not support <files/image>)

end goal for UI/API layer is to limit bad inputs early, and end goal for backend is to gracefully handle responses and emit events in case bad inputs do pass through

this above approach would be much simpler than having to handle it in the core (but core should also have some checks/exceptions/event emitted for this case, because end-users might use our library/sdk directly and not our APIs which would bypass these checks). 

to handle it in the core, i think chat-session.ts where we call the llmservice completeTask is a good place to pre-check the inputs before the input makes it into the llm service. raising exceptions in libraries is the best UX for consumers using our library code (can return UnsupportedInputException or something like that). and this can then be handled in the agent class which would return the answer back to the user. could handle it in agent.run() too, but chat session is likely better because current LLM could vary based on chat session

other edge cases i could think of:

- user on the web UI starts with LLM 1 which supports audio, uploads an audio file, gets the response, then switches to LLM 2, which doesn't support audio sometime in the chat. it's an unlikely edge case, but even with the above changes, the API would randomly start failing which would be unintuitive for the user. (because message history now contains an unsupported format for the new LLM)
- extension of previous one - user switches from LLM 2 back to LLM3 which supports audio. ideally, now the LLM3 should have the audio file context again because this LLM supported it

this leads to a possible solution where we can add a layer that filters the message history getting piped into the LLM based on the currently chosen LLM (somewhere in contextmanager/formatter layer maybe). that way the actual message history still has the audio file input, but we block the audio file from getting sent into LLMs that don't support it, while being able to still send it into LLMs that do support it in case llm switches later on

shaunak99:
Thanks for the detailed feedback! Very insightful for understanding why and how to go about the implementation. 

I've tried to address most of it while reducing as much of the code duplication, and type fixes. Tested it with gemini-2.5-flash.

Check it out and let me know!

src/app/api/server.ts
shaunak99: so the llm apis technically only process text,img,audio at a token level. They added pdf support afaik due to high demand, but internally, they each have a custom preprocessing step internally that either passes the pdf directly as an img or txt+img after OCR or some other file processing technique. 

This is similar to what we do with the file contributor, where we process the different file formats like .txt .md and extract the text (similar to what they do for pdf internally), but we can extend it for more formats like .doc .docx .csv etc. Challenges are what is the best form of data representation for these kinds of docs when you extract the data. Doc/docx sometimes has imgs, diagrams and other unicode. Tabular data has similar challenges.

rahulkarajgikar: i see

we can add a separate issue for this probably since it would require some special handling 

i guess txt and md would be easy but doc formats would have challenges

rahulkarajgikar: added #247

src/app/api/server.ts
```diff
@@ -84,7 +84,73 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
+                const { base64, mimeType, filename } = req.body.fileData;
+
+                // Security validation: file size check (max 50MB for base64)
+                if (typeof base64 === 'string' && base64.length > 67108864) {
+                    // ~50MB in base64
+                    return res.status(400).send({ error: 'File size too large (max 50MB)' });
+                }
+
+                // Security validation: MIME type allowlist
+                const allowedMimeTypes = [
```
rahulkarajgikar: move to ALLOWED_MIME_TYPES constant (coderabbit has similar cmt)

src/core/ai/llm/registry.ts
```diff
@@ -280,6 +347,202 @@ export function isValidProvider(provider: string): boolean {
     return provider.toLowerCase() in LLM_REGISTRY;
 }
 
+/**
+ * Gets the supported file types for a given provider.
+ * @param provider The name of the provider.
+ * @returns Array of supported file types for the provider, or empty array if provider not found.
+ */
+export function getSupportedFileTypes(provider: string): SupportedFileType[] {
```
rahulkarajgikar: i don't think we need provider level functions or defaults

let's keep it at model level only, generalizing across a provider can be unreliable, some models for that provider may support, some models may not.

can remove all the functions that accept only provider and not model

src/app/api/server.ts
```diff
@@ -124,6 +124,29 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
                     }
                 }
 
+                // Model-specific file format validation
+                const currentConfig = agent.getEffectiveConfig(sessionId);
+                const { validateFileForLLM, createFileValidationError } = await import(
+                    '@core/ai/llm/validation.js'
+                );
+
+                const validation = validateFileForLLM(
```
rahulkarajgikar: we can make a more generic function validateInputForLLM - that accepts the current raw inputs and LLM(image, filedata)

that can check for both image supported and file supported

we also don't need the checks from lines 93->126, this is duplicate checks which we can handle once in the validation function

overall api code should be pretty short:

1. extract inputs
2. get current llm config for session
3. validate entire input + llm combination (can delegate to the validate function)
4. return response

similar for APIs below

src/app/webui/components/MessageList.tsx
```diff
@@ -2,7 +2,16 @@
 
 import React, { useEffect, useRef, useState } from 'react';
```
rahulkarajgikar, line 3: not reviewing this file, just trusting that this works

src/app/webui/components/hooks/useChat.ts
```diff
@@ -12,15 +12,61 @@ export interface ImagePart {
...
+    return typeof part === 'object' && part !== null && (part as any).type === 'text';
+}
+
+export function isImagePart(part: unknown): part is ImagePart {
+    return typeof part === 'object' && part !== null && (part as any).type === 'image';
+}
+
+export function isFilePart(part: unknown): part is FilePart {
+    return typeof part === 'object' && part !== null && (part as any).type === 'file';
+}
```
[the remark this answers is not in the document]
rahulkarajgikar: yeah remove any type

src/core/ai/llm/messages/formatters/openai.ts
```diff
@@ -18,9 +23,26 @@ export class OpenAIMessageFormatter implements IMessageFormatter {
      * @param systemPrompt System prompt to include at the beginning of messages
      * @returns Array of messages formatted for OpenAI's API
      */
-    format(history: Readonly<InternalMessage[]>, systemPrompt: string | null): any[] {
+    format(
+        history: Readonly<InternalMessage[]>,
+        systemPrompt: string | null,
+        context?: any
```
rahulkarajgikar: `any` type

src/core/ai/llm/messages/formatters/types.ts
```diff
@@ -1,5 +1,20 @@
 import { InternalMessage } from '../types.js';
 
+/**
+ * Context interface for message formatters.
+ * Provides runtime information for model-aware processing.
+ */
+export interface FormatterContext {
+    /** MCP manager for tool handling */
+    mcpManager?: any; // TODO: Replace with proper MCPManager type when available
```
rahulkarajgikar: we have the mcpManager type

src/core/ai/llm/messages/formatters/types.ts
```diff
@@ -1,5 +1,20 @@
...
+
+    /** LLM provider name (e.g., 'google.generative-ai', 'openai') */
+    llmProvider?: string;
```
rahulkarajgikar: can just stick to provider/model convention (no need for llm prefix - more consistent)

src/core/ai/llm/messages/formatters/vercel.ts
```diff
@@ -23,9 +30,39 @@ export class VercelMessageFormatter implements IMessageFormatter {
      * @param systemPrompt System prompt to include at the beginning of messages
      * @returns Array of messages formatted for Vercel's API
      */
-    format(history: Readonly<InternalMessage[]>, systemPrompt: string | null): any[] {
+    format(
+        history: Readonly<InternalMessage[]>,
+        systemPrompt: string | null,
+        context?: any
```
rahulkarajgikar: no any type

src/core/ai/llm/messages/utils.ts
```diff
@@ -108,3 +125,112 @@ export function getImageData(imagePart: {
...
+    } else if (data instanceof ArrayBuffer) {
+        return Buffer.from(new Uint8Array(data)).toString('base64');
+    } else if (data instanceof URL) {
+        return data.toString();
+    }
+    console.warn('Unexpected file data type in getFileData:', typeof data);
+    return '';
+}
+
+import { validateFileForLLM } from '../validation.js';
```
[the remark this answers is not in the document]
rahulkarajgikar: +1

src/core/ai/llm/messages/formatters/vercel.ts
```diff
@@ -23,9 +30,39 @@ export class VercelMessageFormatter implements IMessageFormatter {
...
+                logger.debug(`Applied Vercel filtering for ${modelInfo}`);
+            } else {
+                // No context available - apply conservative filtering (remove all file parts)
+                logger.warn(
+                    'No model context available for Vercel formatter, applying conservative file filtering'
+                );
+                filteredHistory = filterMessagesByCapabilities([...history], 'groq');
+            }
+        } catch (error) {
+            console.warn('Failed to apply capability filtering, using original history:', error);
```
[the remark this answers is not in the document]
rahulkarajgikar: +1

src/core/ai/llm/messages/manager.ts
```diff
@@ -525,7 +525,7 @@ export class ContextManager {
         try {
             // Use pre-computed system prompt if provided
             const prompt = systemPrompt ?? (await this.getSystemPrompt(context));
-            return this.formatter.format([...messageHistory], prompt);
+            return this.formatter.format([...messageHistory], prompt, context);
```
rahulkarajgikar: trusting that this works cuz we don't have tests, let's just validate this manually

src/core/ai/llm/messages/utils.ts
```diff
@@ -149,3 +149,88 @@ export function getFileData(filePart: {
...
+ * @deprecated Use filterMessagesByLLMCapabilities instead
+ * Filters message content based on specific model capabilities.
+ */
+export function filterMessagesByModelCapabilities(
+    messages: InternalMessage[],
+    provider: string,
+    model: string
+): InternalMessage[] {
+    return filterMessagesByLLMCapabilities(messages, { provider, model });
+}
```
rahulkarajgikar: delete

to avoid this kind of generation, add the prompt 'don't worry about backward compatibility' for PR refactors

src/core/ai/llm/registry.ts
```diff
@@ -37,48 +41,108 @@ export type LLMProvider =
...
+                supportedFileTypes: ['pdf', 'audio'],
+            },
+            { name: 'o4-mini', maxInputTokens: 200000, supportedFileTypes: ['pdf'] },
+            { name: 'o3', maxInputTokens: 200000, supportedFileTypes: ['pdf'] },
+            { name: 'o3-mini', maxInputTokens: 200000, supportedFileTypes: ['pdf'] },
+            { name: 'o1', maxInputTokens: 200000, supportedFileTypes: ['pdf'] },
         ],
         supportedRouters: ['vercel', 'in-built'],
         baseURLSupport: 'none',
+        supportedFileTypes: ['pdf'], // Default for OpenAI models
```
rahulkarajgikar, line 85: let's not keep provider level defaults

we will always know what the current model is at any point in time, so we can have exact checks based on what the llm is

can also enforce that every entry for every LLM in this registry has the `supportedFileTypes` entry

src/core/ai/llm/registry.ts
```diff
@@ -280,6 +347,202 @@ export function isValidProvider(provider: string): boolean {
...
+    const providerInfo = LLM_REGISTRY[normalizedProvider as LLMProvider];
+    return providerInfo ? providerInfo.supportedFileTypes : [];
+}
+
+/**
+ * Normalizes provider names from various sources (e.g., Vercel AI SDK) to our registry keys.
+ * @param provider The provider name to normalize.
+ * @returns The normalized provider name that matches our registry.
+ */
+export function normalizeProviderName(provider: string): string {
```
rahulkarajgikar: don't think we need this, we check inputs way beforehand at the yaml schema, we wouldn't have inputs like `gemini` as the provider at this stage so normalization isn't necessary

src/core/ai/llm/registry.ts
```diff
@@ -280,6 +347,202 @@ export function isValidProvider(provider: string): boolean {
...
+    return modelInfo?.supportedFileTypes ?? providerInfo.supportedFileTypes;
+}
+
+/**
+ * Checks if a provider supports a specific file type.
+ * @param provider The name of the provider.
+ * @param fileType The file type to check support for.
+ * @returns True if the provider supports the file type, false otherwise.
+ */
+export function supportsFileType(provider: string, fileType: SupportedFileType): boolean {
```
rahulkarajgikar: let's get rid of all functions that accept only provider and not model, we can just keep the ones that have provider + model combo

similar for tests

a good extension here would be to also add functions to check for supportsImageType

then we can have the validateInputForLLM function that checks the entire input (image? + file? + text + llm config ) from the APIs and returns supported/unsupported by calling the helper funcs

src/core/ai/llm/validation.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * Validates if a file is supported by the current LLM configuration.
+ * Uses model-specific validation when available, falls back to provider-level.
+ * @param config The LLM configuration containing provider and optional model
+ * @param mimeType The MIME type of the file to validate
+ * @returns Validation result with support status and error details
+ */
+export function validateFileForLLM(
```
rahulkarajgikar: we can make this `validateInputForLLM` as mentioned earlier that accepts the entire API inputs

src/core/ai/llm/messages/utils.ts
```diff
@@ -149,3 +149,88 @@ export function getFileData(filePart: {
...
+
+/**
+ * Filters message content based on LLM capabilities.
+ * Removes unsupported file attachments while preserving supported content.
+ * Uses model-specific validation when available, falls back to provider-level.
+ * @param messages Array of internal messages to filter
+ * @param config The LLM configuration (provider and optional model)
+ * @returns Filtered messages with unsupported content removed
+ */
+export function filterMessagesByLLMCapabilities(
```
rahulkarajgikar, line 167: nice! this is a very useful helper function

let's add tests for this though, it's hard to validate edge cases manually

(combinations of audio, filetype for different models and validate the history that comes back)

src/core/ai/llm/registry.ts
```diff
@@ -280,6 +347,202 @@ export function isValidProvider(provider: string): boolean {
...
+        'audio/mp3': 'audio',
+        'audio/mpeg': 'audio',
+        'audio/wav': 'audio',
+        'audio/x-wav': 'audio',
+        'audio/wave': 'audio',
+        'audio/webm': 'audio',
+        'audio/ogg': 'audio',
+        'audio/m4a': 'audio',
+        'audio/aac': 'audio',
+    };
```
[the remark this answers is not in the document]
rahulkarajgikar: +1, a top level constant in the registry is the only place where we need this list

src/core/ai/session/chat-session.ts
```diff
@@ -240,6 +240,42 @@ export class ChatSession {
         logger.debug(
             `Running session ${this.id} with input: ${input}, imageDataInput: ${imageDataInput}, fileDataInput: ${fileDataInput}`
         );
+
+        // Validate file format support if file data is provided
+        if (fileDataInput) {
```
rahulkarajgikar: with the validateInputForLLM function, this code would also become much better -> we'd just send the whole input + llm combination

we would get back some response, saying which part was supported, which was not, and can emit events accordingly

src/core/ai/systemPrompt/types.ts
```diff
@@ -3,6 +3,9 @@ import { MCPManager } from '../../client/manager.js';
 // Context passed to dynamic contributors
 export interface DynamicContributorContext {
     mcpManager: MCPManager;
+    // Model information for formatters - used for model-aware file filtering
+    llmProvider?: string;
```
rahulkarajgikar, line 7: keep `provider/model` convention without llm prefix

rahulkarajgikar (changes requested):
nice this is better! especially the `filterMessagesByLLMCapabilities` is a super useful utility

with the new suggested utility `validateInputForLLM` and this, we can cover most of the cases pretty well

these 2 methods + unit tests for them will cover most of the important cases

there is still quite a bit of code duplication which i've tried to address as much as possible 

some stuff we can still improve on:
-> add the registry metadata API for querying the llm registry from the web UI
-> then we can decide what user can upload directly at the UI level
-> this has the added benefit that we can add support for way more models in the web UI

can take this as follow up—as long as the web UI is handling unsupported input types gracefully for now that's good enough for this PR

shaunak99:
Updated to `validateInputForLLM` and also added some tests for the input.

Also tried to do some more refactors and removed any types.

Model switching & filtering seems to work alright! 

Screenshot 2025-07-25 at 3 13 38 PM

src/app/api/server.ts
```diff
@@ -430,94 +371,45 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
-                            return;
-                        }
-
-                        fileDataInput = { data: base64, mimeType, filename };
+                        return;
                     }
 
-                    const sessionId = data.sessionId as string | undefined;
                     const stream = data.stream === true; // Extract stream preference, default to false
                     if (imageDataInput) logger.info('Image data included in message.');
```
rahulkarajgikar: nit: move logs in line 414-416 before validation, similar in other apis also
flow would then be -> log the inputs, validate the inputs, return the response

currently if validation fails earlier we wouldn't see logs which could be harder to debug

src/core/ai/llm/messages/formatters/vercel.ts
```diff
@@ -56,10 +55,12 @@ export class VercelMessageFormatter implements IMessageFormatter {
                 logger.warn(
                     'No model context available for Vercel formatter, applying conservative file filtering'
                 );
-                filteredHistory = filterMessagesByCapabilities([...history], 'groq');
+                filteredHistory = filterMessagesByLLMCapabilities([...history], {
+                    provider: 'groq',
```
rahulkarajgikar: why are we using groq?

there shouldn't be an else case here, context.provider and context.model have to be mandatory, otherwise we should throw an error. defaulting to other providers would cause weird behaviours

we should make provider and model non-optional in `FormatterContext` type and any other new types introduced that include provider and model. this would be like a pre-condition for any executions

src/core/ai/llm/messages/formatters/vercel.ts
```diff
@@ -187,10 +190,10 @@ export class VercelMessageFormatter implements IMessageFormatter {
                             if (part.type === 'tool-result') {
                                 let content: InternalMessage['content'];
                                 if (Array.isArray(part.experimental_content)) {
-                                    content = part.experimental_content.map((img: any) => ({
+                                    content = part.experimental_content.map((img: unknown) => ({
                                         type: 'image',
-                                        image: img.data,
-                                        mimeType: img.mimeType,
+                                        image: (img as any).data,
+                                        mimeType: (img as any).mimeType,
```
rahulkarajgikar: nit: llm cursed any kink is back. we might need to use/create an ImageData type for this

src/core/ai/llm/messages/utils.test.ts
```diff
@@ -0,0 +1,322 @@
...
+        const config: FilteringConfig = { provider: 'openai', model: 'gpt-3.5-turbo' };
+
+        const result = filterMessagesByLLMCapabilities(messages, config);
+
+        expect(result[0]!.content).toEqual([
+            { type: 'text', text: '[File attachment removed - not supported by gpt-3.5-turbo]' },
+        ]);
+    });
+
+    test('should add placeholder text without model name when no model specified', () => {
```
rahulkarajgikar: should fail when no model specified, make both provider and model mandatory

src/core/ai/llm/messages/utils.ts
```diff
@@ -184,7 +183,15 @@ export function filterMessagesByLLMCapabilities(
 
                 // Filter file parts based on LLM capabilities
                 if (part.type === 'file' && part.mimeType) {
-                    const validation = validateFileForLLM(config, part.mimeType);
+                    // If no model is specified, reject all files
+                    if (!config.model) {
+                        return false;
+                    }
```
rahulkarajgikar: raise error instead, this is an unexpected case

src/core/ai/llm/messages/utils.ts
```diff
@@ -212,25 +219,4 @@ export function filterMessagesByLLMCapabilities(
...
- * Filters message content based on specific model capabilities.
- */
-export function filterMessagesByModelCapabilities(
-    messages: InternalMessage[],
-    provider: string,
-    model: string
-): InternalMessage[] {
-    return filterMessagesByLLMCapabilities(messages, { provider, model });
-}
+// Removed deprecated filter functions - no backward compatibility needed
```
rahulkarajgikar: stale line

src/core/ai/llm/validation.ts
```diff
@@ -1,62 +1,175 @@
...
+function validateImageInput(
+    _imageData: ImageData,
+    _config: ValidationLLMConfig
+): NonNullable<InputValidationResult['imageValidation']> {
+    // For now, assume images are supported (existing behavior)
+    // This can be expanded later with proper image capability validation
+    return {
+        isSupported: true,
+    };
+}
```
rahulkarajgikar, line 151: let's add an issue for making this change

should be pretty straightforward change, can get community to pick it up

1. update registry to add supportsImage? for each LLM
2. add a function in registry.ts to check the registry
3. update this `validateImageInput` function
4. add tests

src/core/ai/session/chat-session.ts
```diff
@@ -241,40 +242,7 @@ export class ChatSession {
...
-                });
-
-                throw error;
-            }
-
-            logger.debug(
-                `File validation passed for ${validation.fileType} file in session ${this.id}`
-            );
-        }
+        // Note: Input validation is now handled at the API layer before reaching this method
```
rahulkarajgikar: we still need to validate again here, because not all flows will go through API layer (SDK users)

we can re-use the validateInputForLLM.. func here and emit unsupported input events  in case of the validation errors. need to update event bus also to add this as a known event

src/core/ai/llm/validation.ts
```diff
@@ -1,62 +1,175 @@
...
+ * This is the single entry point for all input validation.
+ * @param input The input data to validate (text, image, file)
+ * @param config The LLM configuration (provider and model)
+ * @returns Comprehensive validation result
  */
-export function validateFileForLLM(
-    config: FileValidationConfig,
-    mimeType: string
-): FileValidationResult {
+export function validateInputForLLM(
```
rahulkarajgikar, line 40: `filterMessages` func has good coverage but we need some tests for this func as well as it's a critical point

rahulkarajgikar (changes requested):
changes look mostly good! few minor things still to be addressed, then we should be good

shaunak99:
Addressed most of the comments. Will create the issue for image validation once we merge this.

src/core/ai/llm/messages/manager.ts
```diff
@@ -525,7 +525,17 @@ export class ContextManager {
         try {
             // Use pre-computed system prompt if provided
             const prompt = systemPrompt ?? (await this.getSystemPrompt(context));
-            return this.formatter.format([...messageHistory], prompt, context);
+
+            // Map DynamicContributorContext to FormatterContext
+            const formatterContext: FormatterContext | undefined = context.llmProvider
```
rahulkarajgikar, line 530: nit: we should rechange dynamiccontributor to use provider/model convention. marking for my reference

src/core/ai/llm/validation.test.ts
```diff
@@ -0,0 +1,417 @@
...
+    const mockValidateModelFileSupport = vi.mocked(registry.validateModelFileSupport);
+    const mockGetAllowedMimeTypes = vi.mocked(registry.getAllowedMimeTypes);
+
+    beforeEach(() => {
+        vi.clearAllMocks();
+        // Default mock implementations
+        mockGetAllowedMimeTypes.mockReturnValue(['application/pdf', 'audio/mp3', 'audio/wav']);
+        mockValidateModelFileSupport.mockReturnValue({
+            isSupported: true,
+        });
```
rahulkarajgikar, line 21: @coderabbitai does it make sense to mock these functions - my thought was that we shoudl just use the direct registry functions and only mock for edge-cases where we don't have valid examples, wdyt? mocking these doesn't seem helpful to me (we test mock behaviour rather than real behaviour)

rahulkarajgikar (approved):
validation.test.ts might need some changes, waiting on coderabbit comments

src/core/ai/llm/validation.test.ts
rahulkarajgikar: ok i will handle this in https://github.com/truffle-ai/saiki/pull/249/files, can you suggest the tests

rahulkarajgikar:
merging the PR

pending stuff:

- add issue for image support
- add issue for API to query registry + checking input support directly in web UI

- fix validation.test.ts to remove mocks - i'll handle this in #249 
- update documentation for new events added - i'll handle this in #249

shaunak99:
Great PR! I learned a lot. Will keep in mind for future

PR #260, opened 2025-07-26
Enhance error handling in CLI and registry with custom error classes.…

yanurag-dev:
… Introduce ValidationError, UnknownProviderError, UnknownModelError, and ConfigurationError for improved diagnostics. Update error logging in CLI and registry functions to utilize these new error types. Add tests for new error handling in validation and loader modules.

## Summary by CodeRabbit

* **New Features**
  * Introduced detailed error messages and custom error types for configuration and AI model validation, improving clarity when issues occur.
  * Enhanced API endpoint responses to provide clearer status updates and centralized error handling.
  * Made file input validation function publicly accessible for broader use.

* **Bug Fixes**
  * Improved detection and messaging for unknown providers, models, and configuration file errors.
  * Updated error handling to throw precise exceptions instead of generic errors.

* **Documentation**
  * Updated API type documentation to reflect new error types and naming conventions.

* **Refactor**
  * Standardized error handling throughout the application using custom error classes.
  * Renamed validation error interfaces for consistency across the codebase.
  * Refactored configuration loading and writing with granular error handling and detailed documentation.

* **Tests**
  * Expanded and improved unit tests to cover new error cases and ensure robust error handling.

### DIFF
--- a/src/core/config/loader.test.ts
+++ b/src/core/config/loader.test.ts
@@ -8,8 +8,6 @@ import {
     ConfigFileReadError,
     ConfigParseError,
 } from '@core/error/index.js';
-import dotenv from 'dotenv';
-dotenv.config();
 
 const tmpFile = path.resolve(process.cwd(), 'src/core/config/temp-config.yml');
 


yanurag-dev:
@rahulkarajgikar Pushed some new changes. Pls review it

src/core/ai/llm/validation.test.ts
```diff
@@ -120,7 +120,9 @@ describe('validateInputForLLM', () => {
             );
 
             expect(result.isValid).toBe(false);
-            expect(result.errors).toContain('File size too large (max 50MB)');
+            expect(result.errors.length).toBeGreaterThan(0);
+            expect(result.errors.length).toBeGreaterThan(0);
```
rahulkarajgikar: duplicate

src/core/ai/llm/validation.test.ts
```diff
@@ -137,7 +139,9 @@ describe('validateInputForLLM', () => {
             );
 
             expect(result.isValid).toBe(false);
-            expect(result.errors).toContain('Invalid file data format');
+            expect(result.errors.length).toBeGreaterThan(0);
+            expect(result.errors.length).toBeGreaterThan(0);
```
rahulkarajgikar: duplicate

src/core/config/loader.ts
```diff
@@ -54,13 +55,14 @@ export async function loadAgentConfig(configPath?: string): Promise<AgentConfig>
             const expandedConfig = expandEnvVars(config);
             return expandedConfig;
         } catch (parseError) {
-            throw new Error(
-                `Failed to parse YAML: ${parseError instanceof Error ? parseError.message : String(parseError)}`
+            throw new ConfigurationError(
```
rahulkarajgikar: i'm realizing the existing logic in this function is a bit bad, we have nested try catch blocks

if yaml parsing fails, we catch the error here, then throw it, then catch it again in the top level block - which loses a lot of information 

we should instead break this function up into separate try catch blocks  and introduce specific error classes for each failure point:

ex:

```
Better Approach: Specific Error Classes

  // In src/core/error/index.ts
  export class ConfigFileNotFoundError extends Error {
      constructor(configPath: string) {
          super(`Configuration file not found:
  ${configPath}`);
          this.name = 'ConfigFileNotFoundError';
      }
  }

  export class ConfigFileReadError extends Error {
      constructor(configPath: string, cause: string) {
          super(`Failed to read configuration file
  '${configPath}': ${cause}`);
          this.name = 'ConfigFileReadError';
      }
  }

  export class ConfigParseError extends Error {
      constructor(configPath: string, cause: string) {
          super(`Failed to parse YAML in '${configPath}':
   ${cause}`);
          this.name = 'ConfigParseError';
      }
  }

  export class ConfigEnvVarError extends Error {
      constructor(configPath: string, cause: string) {
          super(`Failed to expand environment variables
  in '${configPath}': ${cause}`);
          this.name = 'ConfigEnvVarError';
      }
  }
```

  Updated loader.ts

```
  export async function loadAgentConfig(configPath?:
  string): Promise<AgentConfig> {
      const absolutePath = resolveConfigPath(configPath);

      // Step 1: Check file exists
      try {
          await fs.access(absolutePath);
      } catch {
          throw new
  ConfigFileNotFoundError(absolutePath);
      }

      // Step 2: Read file
      let fileContent: string;
      try {
          fileContent = await fs.readFile(absolutePath,
  'utf-8');
      } catch (error) {
          throw new ConfigFileReadError(
              absolutePath,
              error instanceof Error ? error.message :
  String(error)
          );
      }

      // Step 3: Parse YAML
      let config: any;
      try {
          config = parseYaml(fileContent);
      } catch (error) {
          throw new ConfigParseError(
              absolutePath,
              error instanceof Error ? error.message :
  String(error)
          );
      }

      // Step 4: Expand environment variables
      try {
          return expandEnvVars(config);
      } catch (error) {
          throw new ConfigEnvVarError(
              absolutePath,
              error instanceof Error ? error.message :
  String(error)
          );
      }
  }
```


we can even go one step further, make each of these error classes related to configuration, extend a common ConfigurationError class. hierarchical exception structure is the best of both worlds:
 -  we still have the specificity of low level errors
 -  we can still catch the high level ConfigurationError in code that calls this function
 -  tests can test for specific exceptions


explained below with sample code + reasons for why this is the best approach:

```
// Parent class for all configuration-related errors
  export abstract class ConfigurationError extends Error
  {
      constructor(
          message: string,
          public readonly configPath?: string
      ) {
          super(message);
          this.name = this.constructor.name; // Auto-sets
   to subclass name
      }
  }

  // Specific configuration error types
  export class ConfigFileNotFoundError extends
  ConfigurationError {
      constructor(configPath: string) {
          super(`Configuration file not found:
  ${configPath}`, configPath);
      }
  }

  export class ConfigFileReadError extends
  ConfigurationError {
      constructor(configPath: string, cause: string) {
          super(`Failed to read configuration file:
  ${cause}`, configPath);
      }
  }

  export class ConfigParseError extends
  ConfigurationError {
      constructor(configPath: string, cause: string) {
          super(`Failed to parse YAML configuration:
  ${cause}`, configPath);
      }
  }

  export class ConfigEnvVarError extends
  ConfigurationError {
      constructor(configPath: string, envVar: string,
  cause?: string) {
          const message = cause
              ? `Failed to expand environment variable
  '${envVar}': ${cause}`
              : `Environment variable '${envVar}' is
  required but not set`;
          super(message, configPath);
      }

      constructor(
          configPath: string,
          public readonly envVar: string,
          cause?: string
      ) {
          // ... message construction
      }
  }

  export class ConfigValidationError extends
  ConfigurationError {
      constructor(
          configPath: string,
          public readonly field: string,
          public readonly validationError: string
      ) {
          super(`Configuration validation failed for
  '${field}': ${validationError}`, configPath);
      }
  }

  Benefits of Hierarchical Structure

  1. Flexible Error Handling

  // CLI can handle all config errors generically
  try {
      const config = await loadAgentConfig();
  } catch (error) {
      if (error instanceof ConfigurationError) {
          logger.error(`Configuration error in
  ${error.configPath}:`);
          logger.error(error.message);

          // But still handle specific cases
          if (error instanceof ConfigFileNotFoundError) {
              logger.info(`Run 'saiki init' to create a
  configuration file`);
          } else if (error instanceof ConfigParseError) {
              logger.info(`Check YAML syntax at
  ${error.configPath}`);
          } else if (error instanceof ConfigEnvVarError)
  {
              logger.info(`Set environment variable:
  ${error.envVar}`);
          }

          process.exit(1);
      }
      // Handle non-config errors differently
      throw error;
  }

  2. Better API Error Responses

  // API can provide consistent error structure
  catch (error) {
      if (error instanceof ConfigurationError) {
          return res.status(400).json({
              error: 'Configuration Error',
              type: error.constructor.name,
              message: error.message,
              configPath: error.configPath,
              // Add specific context based on error type
              ...(error instanceof ConfigEnvVarError && {

                  missingEnvVar: error.envVar
              }),
              ...(error instanceof ConfigValidationError
  && {
                  field: error.field
              })
          });
      }
  }

  3. Better Testing

  // Can test at different levels of specificity
  expect(() =>
  loadConfig('missing.yml')).toThrow(ConfigurationError);
   // Generic
  expect(() => loadConfig('missing.yml')).toThrow(ConfigF
  ileNotFoundError); // Specific

  // Test inheritance
  expect(new ConfigFileNotFoundError('/path')).toBeInstan
  ceOf(ConfigurationError);
  expect(new ConfigParseError('/path', 'bad
  yaml')).toBeInstanceOf(ConfigurationError);

  4. Better Logging/Monitoring

  // Can group all config errors for monitoring
  if (error instanceof ConfigurationError) {
      metrics.increment('config_errors', {
          type: error.constructor.name,
          configPath: error.configPath
      });
  }

  5. Consistent Error Properties

  All config errors automatically have configPath, but
  specific errors can add their own properties:

  // All config errors have configPath
  error.configPath

  // Specific errors have additional context
  (error as ConfigEnvVarError).envVar
  (error as ConfigValidationError).field

```

updated loader.ts

```
export async function loadAgentConfig(configPath?:
  string): Promise<AgentConfig> {
      const absolutePath = resolveConfigPath(configPath);

      // Check file exists
      try {
          await fs.access(absolutePath);
      } catch {
          throw new
  ConfigFileNotFoundError(absolutePath);
      }

      // Read file
      let fileContent: string;
      try {
          fileContent = await fs.readFile(absolutePath,
  'utf-8');
      } catch (error) {
          throw new ConfigFileReadError(
              absolutePath,
              error instanceof Error ? error.message :
  String(error)
          );
      }

      // Parse YAML
      let config: any;
      try {
          config = parseYaml(fileContent);
      } catch (error) {
          throw new ConfigParseError(
              absolutePath,
              error instanceof Error ? error.message :
  String(error)
          );
      }

      // Expand environment variables
      try {
          return expandEnvVars(config);
      } catch (error) {
          throw new ConfigEnvVarError(
              absolutePath,
              'unknown', // Could extract env var name
  from error
              error instanceof Error ? error.message :
  String(error)
          );
      }
  }
```

in general, having more specific error classes is better -> leads you to the buggy code much faster when you see that exception + it is also more informative for users

hierarchical exception also let's calling functions simply check for the parent exception, then can narrow down into child exception classes if necessary

rahulkarajgikar: then we can update tests to check for the new error classes

src/core/ai/llm/validation.test.ts
```diff
@@ -154,9 +158,8 @@ describe('validateInputForLLM', () => {
             );
 
             expect(result.isValid).toBe(false);
-            expect(result.errors).toContain(
-                'Model must be specified for file capability validation'
-            );
+            expect(result.errors.length).toBeGreaterThan(0);
+            expect(result.errors[0]).toContain('Model must be specified');
         });
```
rahulkarajgikar, line 164: let's also add a case for when ValidationError gets thrown

src/core/ai/llm/registry.test.ts
```diff
@@ -276,13 +292,13 @@ describe('File support functionality', () => {
...
             expect(() => getSupportedFileTypesForModel('unknown-provider', 'any-model')).toThrow(
-                "Provider 'unknown-provider' not found in LLM registry"
+                "Provider 'unknown-provider' not found. Available providers: openai, openai-compatible, anthropic, google, groq, xai, cohere"
             );
         });
 
         it('should be case-sensitive for provider names but case-insensitive for model names', () => {
             expect(() => getSupportedFileTypesForModel('OpenAI', 'gpt-4o')).toThrow(
-                "Provider 'OpenAI' not found in LLM registry"
+                "Provider 'OpenAI' not found. Available providers: openai, openai-compatible, anthropic, google, groq, xai, cohere"
```
rahulkarajgikar: since we now have a custom error class, we can change these tests to check for the custom error types instead of specific strings, which is miore robust

some examples:

```
Recommended Test Updates

  it('should throw UnknownProviderError for unknown
  provider', () => {
      expect(() =>
  getSupportedFileTypesForModel('unknown-provider',
  'any-model'))
          .toThrow(UnknownProviderError);
  });

  it('should throw UnknownModelError for unknown model',
  () => {
      expect(() =>
  getSupportedFileTypesForModel('openai',
  'unknown-model'))
          .toThrow(UnknownModelError);
  });

  it('should include available providers in
  UnknownProviderError', () => {
      try {

  getSupportedFileTypesForModel('unknown-provider',
  'any-model');
          fail('Expected UnknownProviderError to be
  thrown');
      } catch (error) {

  expect(error).toBeInstanceOf(UnknownProviderError);
          expect((error as
  UnknownProviderError).message).toContain('Available
  providers:');
      }
  });
```

for any tests we are adding/updating based on the custom exception class, we should follow this approach

checking for thrown error types is much more robust than checking for strings (someone can change the logs or add a new provider, tests break if we have string check)

src/app/api/server.ts
```diff
@@ -110,25 +111,27 @@ export async function initializeApi(agent: SaikiAgent, agentCardOverride?: Parti
...
                 );
+            } catch (error) {
+                if (error instanceof ValidationError) {
+                    return res.status(400).send({
+                        error: error.message,
+                        type: error.name,
+                        field: error.field,
+                    });
+                }
+                throw error; // rethrow unexpected errors
```
rahulkarajgikar: this logic isn't correct

we still need the if(!validation.isValid) check

the ValidationError would only get thrown if the code that validates the input itself fails.

there are essentially 3 main cases we need to handle: 

1. validation.isValid = true -> if user input is correct (should be a 2xx http response)
2. validation.isValid = false -> if user input is bad (should be a 4xx http response indicating user error)
3. ValidationError thrown -> this happens if the function code itself fails at some point - this means server failed (should be a 5xx - indicating server error)


something like this pseudocode for this function:

```
try
     ...... (rest of above function)
     response = validateInputForLLM(....)
     if response is valid -> call the agent.run and return 202 as before
     if response is invalid -> return 400 with the errors similar to what we had before
catch
      if error instance of ValidationError
            return 500 internal server error with the validation error metadata
      else 
            return 500 internal server error 
```

does this make sense?


earlier code more or less handled this, i think all we really need to change is the catch block for if the error is an instance of ValidationError

we need to make similar change in the code below as well

src/core/error/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        this.name = 'UnknownModelError';
+    }
+}
+
+/**
+ * Custom error class for when validation of data fails.
+ * @class ValidationError
+ * @extends {Error}
+ */
+export class ValidationError extends Error {
```
rahulkarajgikar: i think we can name this something more descriptive, ValidationError is not very meaningful. the above exceptions have much better names

maybe `LLMInputValidationError`

src/core/error/index.ts
```diff
@@ -0,0 +1,92 @@
...
+        this.name = 'ValidationError';
+    }
+}
+
+/**
+ * Custom error class for issues related to configuration.
+ * @class ConfigurationError
+ * @extends {Error}
+ */
+export class ConfigurationError extends Error {
```
rahulkarajgikar: explained about this above

rahulkarajgikar (changes requested):

yanurag-dev:
@rahulkarajgikar I have pushed new changes can you pls review it.

src/core/ai/llm/registry.test.ts
```diff
@@ -322,10 +349,17 @@ describe('File support functionality', () => {
...
-                    getSupportedFileTypesForModel('OpenAI', 'gpt-4o-audio-preview')
-                ).toThrow("Provider 'OpenAI' not found in LLM registry");
+            it('should throw UnknownProviderError and include available providers for case-sensitive provider names', () => {
+                const unknownProvider = 'OpenAI';
+                try {
+                    getSupportedFileTypesForModel(unknownProvider, 'gpt-4o-audio-preview');
+                    fail('Expected UnknownProviderError to be thrown');
+                } catch (error) {
+                    expect(error).toBeInstanceOf(UnknownProviderError);
+                }
```
rahulkarajgikar: use expect(func()).toThrow instead of try catch block (like we've done below)

src/core/ai/llm/validation.test.ts
```diff
@@ -412,4 +414,52 @@ describe('createInputValidationError', () => {
             },
         });
     });
+
+    test('should correctly format an error originating from LLMInputValidationError', () => {
+        const validation = {
+            isValid: false,
+            errors: [new LLMInputValidationError('Input text too short', 'text').message], // Simulate error message from LLMInputValidationError
```
rahulkarajgikar, line 446: LLMInputValidationError would never show up in this `validation` result.
It only ever gets _thrown_ by the `validateInputForLLM` function

this test should actually be something like this:

```
test('validateInputForLLM throws LLMInputValidationError if some function fails)
     mock some function called inside validateInputForLLM, say `validateFileInput` to always throw an error
     call validateInputForLLM with some dummy arguments
     expect it to throw the LLMInputValidationError
```

and we can remove the second test `test('should combine multiple error messages, including those from ValidationErrors'`

src/core/ai/session/chat-session.ts
```diff
@@ -244,34 +245,39 @@ export class ChatSession {
...
+                this.eventBus.emit('saiki:unsupportedInput', {
+                    sessionId: this.id,
+                    error: error.message,
+                    type: error.name,
+                });
+                logger.error(`Session ${this.id}: Input validation failed`, {
+                    error: error.message,
+                    type: error.name,
+                    field: error.field,
+                });
```
rahulkarajgikar: this was earlier `llmService:unsupportedInput`, please revert to this input with the same format otherwise this is a breaking change + we also don't need to track session id i had mentioned in previous review

i think we don't need to make any changes in this file, let's revert these changes

src/core/config/loader.test.ts
```diff
@@ -33,42 +38,60 @@ describe('loadAgentConfig', () => {
...
             expect(config.llm).toBeDefined();
         } catch (error) {
-            // If no default config is found, it should throw a specific error
-            expect(error).toBeInstanceOf(Error);
-            expect((error as Error).message).toMatch(/No agent\.yml found|No bundled config found/);
+            expect(error).toBeInstanceOf(ConfigFileNotFoundError);
+            expect((error as ConfigFileNotFoundError).message).toMatch(
+                /Configuration file not found/
+            );
         }
```
rahulkarajgikar, line 96: nice, tests in this file are really good

we can add one more test for the ConfigEnvVarError to cover all the cases

src/core/ai/session/chat-session.ts
rahulkarajgikar: I think there is a bit of confusion about `LLMInputValidationError` and when this gets generated

re-read the `validateInputForLLM` function once

essentially, in this function, we are using normal errors included in the return type response in a regular case. (code runs successfully, but input has some issues(

the reason for this is because it allows us to explain all of the issues with the input easily to the user at once.

however, `LLMInputValidationError`, firstly only ever gets thrown, not included in the return type. And it also only gets thrown in rare case where the code execution itself fails. This will in practice, rarely ever happen.

src/core/config/loader.test.ts
yanurag-dev: "@rahulkarajgikar I have incorporated the requested changes.

Just need your help here to test ConfigEnvVarError.

This error should be thrown if there is an issue with the expansion of an environment variable or if a variable is missing.

However, the current expandEnvVars function does not throw any error; instead, it returns an empty string.
Should I handle the error-throwing logic inside expandEnvVars?"

rahulkarajgikar: nice catch!

yeah `expandEnvVars` should throw some error, if any env var we see defined in the config isn't defined, then app would have confusing runtime errors, so it is better to catch this beforehand

we should update expandEnvVars code to look something like this:


original:

```
const expanded = config.replace(
            /\$([A-Z_][A-Z0-9_]*)|\${([A-Z_][A-Z0-9_]*)}/gi,
            (_, v1, v2) => {
                return process.env[v1 || v2] || '';
            }
        );
```

new:

```
const expanded = config.replace(
            /\$([A-Z_][A-Z0-9_]*)|\${([A-Z_][A-Z0-9_]*)}/gi,
            (_, v1, v2) => {
                const varName = v1 || v2;
                const value = process.env[varName];
                
                if (value === undefined || value === '') {
                    throw new ConfigEnvVarError(`Environment variable '${varName}' is not defined`);
                }
            }
        );
```

then we can update the loadAgentConfig to rethrow this type of error

so then for the testing part, we can actually just test the expandEnvVars function instead

yanurag-dev:
@rahulkarajgikar I have pushed some changes. can you review the pr pls. 

I have made change in the expand `expandEnvVars` which affected one the testcase in `loader.test.ts`.  Since we are not passing the possible env variable which is leading to through `ConfigEnvVarError` while we are expecting `ConfigFileNotFoundError`. So handled both the error in the testcase. 

If you have any suggestion please let me know.

rahulkarajgikar:
nice! looks good to me, will merge! thanks for this!

PR #275, opened 2025-08-05
Initial implementation of an agent registry

shaunak99:
This PR introduces an initial implementation for an agent registry to enable easy sharing and usage of agent configurations.


src/core/agent-registry/agent-registry.ts
```diff
@@ -0,0 +1,299 @@
...
+import {
+    AgentRegistry,
+    AgentRegistryEntry,
+    AgentRegistryConfig,
+    AgentRegistryConfigSchema,
+} from './types.js';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const PROJECT_ROOT = path.resolve(__dirname, '../../../');
```
rahulkarajgikar: this is brittle, we should have better path resolution logic

i am not sure this will work outside the saiki repo. it will be dependent on where the user runs the installed command from, and will probably fail

relative paths are bad and we should use some walk up logic like we have in path utils

the right way is to call some function that walks up and finds the root directory we need to walk down from

then from there, use the relative paths to walk down and find the registry.json

it should handle the following cases:
1. developers working in saiki repository
2. installed CLI (test with npm pack and npm install)
3. (longer term) inside a saiki initted project (can skip this case for now)

- i'm working on path v2 utils for these cases so we could for now just make sure 1 and 2 work properly (current path utils should be able to do this, i think getSaikiPath handles this - check logging and default agent config logic)

src/core/agent-registry/agent-registry.ts
```diff
@@ -0,0 +1,299 @@
...
+    name: string;
+    displayName: string;
+    description: string;
+    version: string;
+    author?: string;
+    tags?: string[];
+    configUrl: string;
+    lastUpdated?: string;
+}
+
```
rahulkarajgikar: we don't need this interface, we can use types derived from zod schema

src/core/agent-registry/types.ts
```diff
@@ -0,0 +1,57 @@
...
+        tags: z.array(z.string()).optional().describe('Tags for categorization'),
+        configUrl: z.string().describe('URL or local path to the agent configuration'),
+        source: z
+            .enum(['registry', 'remote', 'local'])
+            .describe('Whether this is a registry agent or external'),
+        lastUpdated: z.string().optional().describe('Last update timestamp'),
+    })
+    .strict();
+
+export type AgentRegistryEntry = z.infer<typeof AgentRegistryEntrySchema>;
```
rahulkarajgikar: use z.output instead of z.infer (equivalent, easier to understand)

src/app/index.ts
```diff
@@ -217,6 +262,8 @@ program
             'build complex AI applications like Cursor, and more.\n\n' +
             // TODO: Add `saiki tell me about your cli` starter prompt
             'Run saiki interactive CLI with `saiki` or run a one-shot prompt with `saiki <prompt>`\n' +
+            'Use available agents: `saiki github-agent`, `saiki database-agent`, etc.\n' +
```
rahulkarajgikar: let's stick to using -a or --agent for any agents - without argument it gets confusing because we already have headless commands 

this should also simplify some of the other code because we don't have to check for both cases

src/core/agent-registry/types.ts
```diff
@@ -0,0 +1,57 @@
...
+            .default({}),
+        remoteRegistries: z
+            .array(z.string())
+            .optional()
+            .describe('Remote registry endpoints (for future use)'),
+        cacheTtl: z
+            .number()
+            .positive()
+            .describe('Cache TTL for remote agents (in seconds)')
+            .default(3600),
```
rahulkarajgikar: let's remove remoteRegistries, can add it back later when we support this feature and know how we want to implement this. for now we can just keep the ones we need

src/core/agent-registry/types.ts
```diff
@@ -0,0 +1,57 @@
...
+    })
+    .strict();
+
+export type AgentRegistryConfig = z.infer<typeof AgentRegistryConfigSchema>;
```
rahulkarajgikar: use z.output

src/core/agent-registry/types.ts
```diff
@@ -0,0 +1,57 @@
...
+    .object({
+        name: z.string().describe('Unique identifier for the agent'),
+        displayName: z.string().describe('Human-readable display name'),
+        description: z.string().describe("Brief description of the agent's purpose"),
+        version: z.string().describe('Agent version'),
+        author: z.string().optional().describe('Author/maintainer information'),
+        tags: z.array(z.string()).optional().describe('Tags for categorization'),
+        configUrl: z.string().describe('URL or local path to the agent configuration'),
+        source: z
+            .enum(['registry', 'remote', 'local'])
```
rahulkarajgikar: let's keep this minimal for now with only ones we do support

src/core/agent-registry/agent-registry.ts
```diff
@@ -0,0 +1,299 @@
...
+    async resolveAgent(nameOrPath: string): Promise<string> {
+        logger.debug(`Resolving agent: ${nameOrPath}`);
+
+        // 1. Check if it's a registry agent name
+        if (this.isRegistryAgent(nameOrPath)) {
+            const agent = this.getRegistryAgent(nameOrPath);
+            if (agent) {
+                // If it's an HTTP URL, download and cache it
+                if (agent.configUrl.startsWith('http')) {
+                    const cachedPath = await this.downloadAndCacheConfig(agent.configUrl);
```
rahulkarajgikar: using http paths like this isn't ideal for a few reasons:
1. breaks versioning - the agent configs would be able to change independently of the installed version of the registry which could lead to random things breaking for end users
2. goes against the LocalRegistry idea, which shouldn't fetch from a URL

this is better suited for a remote registry we host where we can handle these cases properly

simpler way for local registry is to use some path based logic to get yml file (either from installed node modules agents directory, or from local folder), then call loadAgentConfig with the filepath

we are already shipping some parts of the  agents directory as part of our distro, and we have logic to read the default agent.yml as well which handles the different cases (inside the repo, globally installed case). we also use similar logic for logging (i think the function is called getSaikiPath())
 
we can probably re-use a lot of this, need to read that logic and see how that works

for any of the new registry.json entries we ship, we just need to also ensure to ship the relevant agents/<yml> file in the distro

we would just need to modify that slightly to make it a bit more generic to read anything from the agents directory.

then our agent.configURL can either be something like `agents/<yml>` or we assume `agents` directory and just make it `<yml>`

maybe for local agents, we rename it to `agent.configFile` instead

then the registry can use this utility function to find the parent folder, and walk down from there to find the relevant yml file

src/core/config/loader.ts
```diff
@@ -69,10 +70,24 @@ function expandEnvVars(config: any): any {
...
+            // If agent registry resolution fails, fall back to the old behavior
+            logger.debug(
+                `Agent registry resolution failed for '${configPath}': ${error instanceof Error ? error.message : String(error)}`
+            );
+            absolutePath = resolveConfigPath(configPath);
+        }
+    } else {
+        // No configPath provided, use default behavior
+        absolutePath = resolveConfigPath(configPath);
+    }
```
rahulkarajgikar, line 54: here, loadAgentConfig role gets confusing because we are handling both config Paths and agent names in one configPath variable


sample flow:

current: loadAgentConfig(configPath) -> does everything, config path could be just a name or a path

proposed new:
- use a util function to find if a string is a path or not( we can use regex to search for characters like . and /, there are probably standard utils for this)
- if it's a path -> call loadAgentConfigFromPath -(equiv to previous loadAgentConifg)
- if it's a name -> check loadAgentConfigFromRegistry -> new function (equiv to current resolveAgentConfig)
- can even have a wrapper function on top loadAgentConfig(nameOrPath) that does all these things and we call that from the top.

then the registry can only accept raw names (and not paths)

src/core/agent-registry/agent-registry.ts
```diff
@@ -0,0 +1,299 @@
...
+                if (rawAgent.author) {
+                    entry.author = rawAgent.author;
+                }
+                if (rawAgent.tags) {
+                    entry.tags = rawAgent.tags;
+                }
+                if (rawAgent.lastUpdated) {
+                    entry.lastUpdated = rawAgent.lastUpdated;
+                }
+
```
rahulkarajgikar: we should use zod for all of this data parsing

update AgentRegistryEntrySchema to have a refine/superRefine block that handles this data transformation

then this loop just becomes:
```
for (const [key, rawAgent] of Object.entries(rawAgents)) {
    agentEntry = AgentRegistryEntrySchema.parse(rawAgent)
     agents[key] = agentEntry
}
```

also don't use a try catch block here - better to raise failures early and not silently skip so we know immediately when something is broken

you can check other schemas for good examples of how to define business logic in zod

shaunak99:
@coderabbitai review

PR #289, opened 2025-08-17
Adding a ResourceManager

shaunak99:
This PR leverages the resource primitive from MCPServers and introduces a new ResourceManager to enable using resources within the application.

The core idea of resources is to enable pulling data/information based on the paths and references provided by the MCPServer. We extended it primarily in our client to enable referencing it as context for LLM to use when referenced by the user but can be extended to other app functionalities as well. 

We also add functionality to reference resources in the webui using `@` for MCPServers that do expose resources. 

Screenshot 2025-08-17 at 7 55 00 PM

## Summary by CodeRabbit

- **New Features**
  - Resource management: list/read resources via new "dexto resources" CLI (supports --read and --json).
  - Chat input: @-triggered resource autocomplete and automatic expansion of @ references into resource content.
  - MCP tools: richer tool metadata shown (id, name, description, input schema).
  - Internal filesystem resources support with sample default paths.

- **Improvements**
  - Standardized error handling for MCP and webhook flows.

- **Tests**
  - New unit tests for resource parsing, manager, schemas, and lifecycle.


shaunak99:
So the resource primitive in MCP is actually derived from standard app development patterns which use URI based address referencing for data or states. This is also similar to how the app design would work internally within Cursor or Slack or any other UI which offers `@mention` kind of syntax for referencing. MCP basically has used this same pattern to make it easy to access over a client-server architecture. 

Having a ResourceManager would be useful interface if we want to extend our system to offer the same custom URI or local data referencing. For eg, our FileContributor right now adds context by default when we enable it but we could repurpose the same function to work via the ResourceManager to reference files that we have made available to the agent, but not autofill in context. Since we are adding and do plan on adding file/data/context referencing capabilities, like our `FileContributor`, this would require some client side implementation of this even if it is not coming from MCP. Even without MCP, you would implement something similar on the client-side which effectively performs the same set of functions for data accessibility and fetching and since we don't have any setup for this currently, this is probably a good place to start and extend. This would also be useful for managing a lot of RAG based systems to reference what data sources are available in our system, even non-MCP ones. We can think about the nomenclature but I think it would be useful to have this, although can cut down on too much many extensibility features. 

What we could do instead though is change it from being "custom" to "internal", similar to how we implemented it for tools in case that simplifies things. But having a client side manager to me makes a lot of sense as it would basically become our core service for managing all data availability and referencing capabilities. The `@` is just one UI implementation of it but you could incorporate it at the app layer in different ways (like referencing a URL to fetch data at runtime similar to cursor). Moving this to become a internal specific resource or setups might be better since we control the functionalities offered similar to internal-tools. I can also add local file referencing (making all files in cwd accessible as referencable files) as a first implementation of this. Can extend to data fetching as needed. 

The next PR after this also supports /prompts which has a similar setup so I can update that to internal prompts as well - this internal-prompts would essentially be the way to create "prompt packs". 

Let me know your thoughts!

PR #306, opened 2025-08-29
[WIP] Patch for media response context explosion

shaunak99:
This patch addresses a bug where toolResponses containing media type data as base64 strings leads to exploded contexts due to addition into the context window as text parts. Currently, we implement a defensive approach and return media types for UI rendering while filtering it completely in the context.

Ideally, the media type should be parsed and filtered based on the model capabilities and added to the context with appropriate text/image/audio/file part. 

Note: Only OpenAI/Anthropic services are addressed in this since Vercel automatically re-ingests toolResponses and needs some explicit work around to prevent this. - Needs to be addressed asap or else token cost can explode very easily.

### DIFF
--- a/src/core/context/manager.ts
+++ b/src/core/context/manager.ts
@@ -6,7 +6,7 @@ import { ICompressionStrategy } from './compression/types.js';
 import { MiddleRemovalStrategy } from './compression/middle-removal.js';
 import { OldestRemovalStrategy } from './compression/oldest-removal.js';
 import { logger } from '../logger/index.js';
-import { getImageData, countMessagesTokens } from './utils.js';
+import { countMessagesTokens, sanitizeToolResultToContent } from './utils.js';
 import { DynamicContributorContext } from '../systemPrompt/types.js';
 import { PromptManager } from '../systemPrompt/manager.js';
 import { IConversationHistoryProvider } from '@core/session/history/types.js';
@@ -421,30 +421,30 @@ export class ContextManager<TMessage = unknown> {
         if (!toolCallId || !name) {
             throw ContextError.toolCallIdNameRequired();
         }
-
-        // Simplest image detection: if result has an 'image' field, treat as ImagePart
+        // Sanitize tool result to avoid adding non-text data as raw text
+        // and to convert media/data-uris/base64 to structured parts.
         let content: InternalMessage['content'];
-        if (result && typeof result === 'object' && 'image' in result) {
-            // Use shared helper to get base64/URL
-            const imagePart = result as {
-                image: string | Uint8Array | Buffer | ArrayBuffer | URL;
-                mimeType?: string;
-            };
-            content = [
-                {
-                    type: 'image',
-                    image: getImageData(imagePart),
-                    mimeType: imagePart.mimeType || 'image/jpeg',
-                },
-            ];
-        } else if (typeof result === 'string') {
-            content = result;
-        } else if (Array.isArray(result)) {
-            // Assume array of parts already
-            content = result;
-        } else {
-            // Fallback: stringify all other values
-            content = JSON.stringify(result ?? '');
+        content = sanitizeToolResultToContent(result);
+
+        // Log what we are storing (brief)
+        if (typeof content === 'string') {
+            const preview = content.slice(0, 200);
+            logger.debug(
+                `ContextManager: Storing tool result (text) for ${name} (len=${content.length}): ${preview}${
+                    content.length > 200 ? '...' : ''
+                }`
+            );
+        } else if (Array.isArray(content)) {
+            const summary = content
+                .map((p) =>
+                    p.type === 'text'
+                        ? `text(${p.text.length})`
+                        : p.type === 'image'
+                          ? `image(${p.mimeType || 'image'})`
+                          : `file(${p.mimeType || 'file'})`
+                )
+                .join(', ');
+            logger.debug(`ContextManager: Storing tool result (parts) for ${name}: [${summary}]`);
         }
 
         await this.addMessage({ role: 'tool', content, toolCallId, name });
--- a/src/core/context/utils.ts
+++ b/src/core/context/utils.ts
@@ -5,8 +5,10 @@ import { validateModelFileSupport } from '@core/llm/registry.js';
 import { LLMContext } from '@core/llm/types.js';
 import { ContextError } from './errors.js';
 
-// Approximation for message format overhead
-const DEFAULT_OVERHEAD_PER_MESSAGE = 4;
+// Tunable heuristics and shared constants
+const DEFAULT_OVERHEAD_PER_MESSAGE = 4; // Approximation for message format overhead
+const MIN_BASE64_HEURISTIC_LENGTH = 512; // Below this length, treat as regular text
+const MAX_TOOL_TEXT_CHARS = 8000; // Truncate overly long tool text
 
 /**
  * Counts the total tokens in an array of InternalMessages using a provided tokenizer.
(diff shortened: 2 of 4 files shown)


tts_Hello_20250802_025327.mp3
rahulkarajgikar, line 1: delete this?

src/core/llm/services/vercel.ts
```diff
@@ -38,6 +39,8 @@ export class VercelLLMService implements ILLMService {
     private sessionEventBus: SessionEventBus;
     private readonly sessionId: string;
     private toolSupportCache: Map<string, boolean> = new Map();
+    // Map of toolCallId -> queue of raw results to emit with that callId
+    private rawResultsByCallId: Map<string, any[]> = new Map();
```
rahulkarajgikar: use unknown not any

or better, get the RESULT type mentioned here: https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#tools.tool.execute

src/core/llm/services/vercel.ts
```diff
@@ -92,14 +95,53 @@ export class VercelLLMService implements ILLMService {
             if (tool) {
                 acc[toolName] = {
                     inputSchema: jsonSchema(tool.parameters),
-                    execute: async (args: unknown) => {
+                    execute: async (args: unknown, _options?: { toolCallId?: string }) => {
```
rahulkarajgikar: don't use underscore for used variables

src/core/llm/services/vercel.ts
```diff
@@ -92,14 +95,53 @@ export class VercelLLMService implements ILLMService {
...
-                            return await this.toolManager.executeTool(
+                            const rawResult = await this.toolManager.executeTool(
                                 toolName,
                                 args as Record<string, unknown>,
                                 this.sessionId
                             );
+
+                            // Queue raw, unfiltered result under the specific toolCallId
+                            const callId = _options?.toolCallId ?? this.sessionId;
+                            const queue = this.rawResultsByCallId.get(callId) ?? [];
```
rahulkarajgikar: don't fallback call id to session id, make the toolCallId non optional (vercel always provides it)

and use that

https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text

src/core/llm/services/vercel.ts
```diff
@@ -92,14 +95,53 @@ export class VercelLLMService implements ILLMService {
...
+                                    asString.length > MAX_LEN
+                                        ? `${asString.slice(0, MAX_LEN)}... [truncated]`
+                                        : asString;
+                                logger.warn(
+                                    `Vercel tool result sanitization failed for ${toolName}: ${String(
+                                        sanErr
+                                    )}. Returning truncated text.`
+                                );
+                                return truncated;
+                            }
```
rahulkarajgikar: we already handle all of this error stuff in the helper functions so we don't need this catch

`sanitizeToolResultToContent` is already a safe function that catches errors, and handles them safely like this, so it never errors out so we don't need this

we can make the summarize func also safe in simialr manner

src/core/llm/services/vercel.ts
```diff
@@ -298,12 +339,22 @@ export class VercelLLMService implements ILLMService {
                             });
                         }
                     }
+                    // Emit tool results: prefer raw mapped by callId; fallback to sanitized output
                     if (step.toolResults && step.toolResults.length > 0) {
                         for (const toolResult of step.toolResults) {
+                            const callId = toolResult.toolCallId;
+                            const sanitized = toolResult.output;
+                            let raw: any | undefined;
```
rahulkarajgikar: don't use any type, similar to above (use REsult type or unknown)

src/core/context/manager.ts
```diff
@@ -432,30 +432,30 @@ export class ContextManager<TMessage = unknown> {
...
-            ];
-        } else if (typeof result === 'string') {
-            content = result;
-        } else if (Array.isArray(result)) {
-            // Assume array of parts already
-            content = result;
-        } else {
-            // Fallback: stringify all other values
-            content = JSON.stringify(result ?? '');
+        content = sanitizeToolResultToContent(result);
```
rahulkarajgikar, line 448: currently vercel vs anthropic/openai handle getting llm responses quite differently

(anthropic/openai add the tool responses, assistant resposnes using the individual methods, while vercel uses processLLMResponse, processStreamResponse.) - let's add A TODO at the top of context manager file to unify these approaches so it's more uniform everywhere and easier to make changes in 1 place for all

src/core/context/utils.ts
```diff
@@ -206,7 +206,295 @@ export function filterMessagesByLLMCapabilities(
...
+    return obj;
+}
+
+/**
+ * Convert an arbitrary tool result into safe InternalMessage content.
+ * - Converts data URIs and base64 blobs to media/file parts
+ * - Removes huge binary blobs inside objects
+ * - Truncates extremely long raw text
+ */
+export function sanitizeToolResultToContent(result: unknown): InternalMessage['content'] {
```
rahulkarajgikar, line 296: add unit tests for this function for known result types to validate it works properly

src/core/context/utils.ts
```diff
@@ -206,7 +206,295 @@ export function filterMessagesByLLMCapabilities(
...
+                return `${head}\n... [${result.length - 5000} chars omitted] ...\n${tail}`;
+            }
+            return result;
+        }
+
+        // Case 2: array of parts or mixed
+        if (Array.isArray(result)) {
+            // Ensure only supported part types (text|image|file) appear in the array
+            const parts: Array<TextPart | ImagePart | FilePart> = [];
+            for (const item of result as any[]) {
```
rahulkarajgikar: don't use any, use unknown and add checks for types if needed

src/core/context/utils.ts
```diff
@@ -206,7 +206,295 @@ export function filterMessagesByLLMCapabilities(
...
+                        });
+                        continue;
+                    }
+                    if ('image' in obj) {
+                        parts.push({
+                            type: 'image',
+                            image: getImageData({ image: obj.image }),
+                            mimeType: obj.mimeType || 'image/jpeg',
+                        });
+                        continue;
```
rahulkarajgikar: can combine with prev if

src/core/context/utils.ts
```diff
@@ -206,7 +206,295 @@ export function filterMessagesByLLMCapabilities(
...
+    return Math.floor((charLength * 3) / 4);
+}
+
+/**
+ * Convert arbitrary tool content to safe text for providers that only accept textual tool messages.
+ * - If content is an array of parts, summarize it.
+ * - If content is a string that looks like base64/data URI, replace with a short placeholder.
+ * - Otherwise pass text through.
+ */
+export function toTextForToolMessage(content: InternalMessage['content'] | string | null): string {
```
rahulkarajgikar: `| string | null` are already covered in `InternalMessage['content']` so this is redundant, we can just directly use `content: InternalMessage['content']`

src/core/context/utils.ts
```diff
@@ -206,7 +206,295 @@ export function filterMessagesByLLMCapabilities(
...
+    if (!b64Regex.test(value)) return false;
+    // Low whitespace / punctuation typical for base64
+    const nonWordRatio = (value.match(/[^A-Za-z0-9+/=]/g)?.length || 0) / value.length;
+    return nonWordRatio < 0.01;
+}
+
+/**
+ * Parse data URI and return { mediaType, base64 } or null if not a data URI.
+ */
+export function parseDataUri(value: string): { mediaType: string; base64: string } | null {
```
rahulkarajgikar, line 251: add few unit tests to validate base64 truncation/etc. properly

rahulkarajgikar (changes requested):

src/core/llm/services/vercel.ts
shaunak99: using unknown for now

shaunak99:
Avoiding these comments for now since a lot of the sanitization code will be re-written when we add FileStorage+hooks and route everything through that

rahulkarajgikar (commented):
alright, let's just test out once and see if the base64 data is flooding UI. 

if yeah, we can temporarily do coderabbit's changes to avoid this, and do a better fix later 

otherwise this is cool to merge

shaunak99:
So we do want the UI to get the raw base64 - this is how the UI renders the image/audio

However, logging the raw was definitely problematic so I have updated that. Works as expected.

Also updated `copy-webui-dist.ts` to patch the webui build issue

PR #310, opened 2025-08-31
Adding support to cancel/interrupt a message

shaunak99:
## Summary by CodeRabbit

* **New Features**
  * Cancel in-progress AI runs:
    - Web UI: Stop button and Esc-to-cancel while a run is processing; UI exposes live "processing" state and cancel action.
    - CLI: Press Esc to cancel without exiting.
    - API: POST cancel endpoint and WebSocket cancel messages to stop current or session-specific runs.
    - Agent/session-level cancel APIs.

* **Improvements**
  * End-to-end cancellation via AbortSignal propagated to LLM providers (streaming & non‑streaming).
  * Suppresses noisy error banners for user-initiated cancels.
  * Optional session get without restoring from storage to avoid I/O.

* **Tests**
  * Updated tests to expect AbortSignal passed to LLM calls.

### DIFF
--- a/src/app/api/server.ts
+++ b/src/app/api/server.ts
@@ -183,6 +183,20 @@ export async function initializeApi(agent: DextoAgent, agentCardOverride?: Parti
         }
     });
 
+    // Cancel an in-flight run for a session (or default session)
+    app.post('/api/cancel', express.json(), async (req, res) => {
+        const sessionId =
+            req.body && typeof req.body.sessionId === 'string' ? req.body.sessionId : undefined;
+        const cancelled = agent.cancel(sessionId);
+        return res.status(200).json({ cancelled, sessionId: sessionId ?? 'default' });
+    });
+
+    app.post('/api/sessions/:sessionId/cancel', async (req, res) => {
+        const { sessionId } = req.params;
+        const cancelled = agent.cancel(sessionId);
+        return res.status(200).json({ cancelled, sessionId });
+    });
+
     // Synchronous endpoint: await the full AI response and return it in one go
     app.post('/api/message-sync', express.json(), async (req, res, next) => {
         logger.info('Received message via POST /api/message-sync');
@@ -443,6 +457,15 @@ export async function initializeApi(agent: DextoAgent, agentCardOverride?: Parti
                         `Processing reset command from WebSocket${sessionId ? ` for session: ${sessionId}` : ''}.`
                     );
                     await agent.resetConversation(sessionId);
+                } else if (data.type === 'cancel') {
+                    const sessionId = data.sessionId as string | undefined;
+                    logger.info(
+                        `Processing cancel command from WebSocket${sessionId ? ` for session: ${sessionId}` : ''}.`
+                    );
+                    const cancelled = agent.cancel(sessionId);
+                    if (!cancelled) {
+                        logger.debug('No in-flight run to cancel');
+                    }
                 } else {
                     logger.warn(`Received unknown WebSocket message type: ${data.type}`);
                     sendWebSocketValidationError(ws, 'Unknown message type', {
--- a/src/app/api/server.ts
+++ b/src/app/api/server.ts
@@ -188,12 +188,18 @@ export async function initializeApi(agent: DextoAgent, agentCardOverride?: Parti
         const sessionId =
             req.body && typeof req.body.sessionId === 'string' ? req.body.sessionId : undefined;
         const cancelled = agent.cancel(sessionId);
-        return res.status(200).json({ cancelled, sessionId: sessionId ?? 'default' });
+        if (!cancelled) {
+            logger.debug(`No in-flight run to cancel for session: ${sessionId || 'default'}`);
+        }
+        return res.status(200).json({ cancelled, sessionId: sessionId || 'default' });
     });
 
     app.post('/api/sessions/:sessionId/cancel', async (req, res) => {
         const { sessionId } = req.params;
         const cancelled = agent.cancel(sessionId);
+        if (!cancelled) {
+            logger.debug(`No in-flight run to cancel for session: ${sessionId}`);
+        }
         return res.status(200).json({ cancelled, sessionId });
     });
 
--- a/src/app/api/server.ts
+++ b/src/app/api/server.ts
@@ -110,6 +110,11 @@ const SearchQuerySchema = z.object({
     role: z.enum(['user', 'assistant', 'system', 'tool']).optional(),
 });
 
+// Schema for cancel request parameters
+const CancelRequestSchema = z.object({
+    sessionId: z.string().min(1, 'Session ID is required'),
+});
+
 // Helper to parse and validate request body
 function parseBody<T>(schema: z.ZodSchema<T>, body: unknown): T {
     return schema.parse(body); // ZodError handled by error middleware
@@ -183,24 +188,18 @@ export async function initializeApi(agent: DextoAgent, agentCardOverride?: Parti
         }
     });
 
-    // Cancel an in-flight run for a session (or default session)
-    app.post('/api/cancel', express.json(), async (req, res) => {
-        const sessionId =
-            req.body && typeof req.body.sessionId === 'string' ? req.body.sessionId : undefined;
-        const cancelled = agent.cancel(sessionId);
-        if (!cancelled) {
-            logger.debug(`No in-flight run to cancel for session: ${sessionId || 'default'}`);
+    // Cancel an in-flight run for a session
+    app.post('/api/sessions/:sessionId/cancel', async (req, res, next) => {
+        try {
+            const { sessionId } = parseQuery(CancelRequestSchema, req.params);
+            const cancelled = await agent.cancel(sessionId);
+            if (!cancelled) {
+                logger.debug(`No in-flight run to cancel for session: ${sessionId}`);
+            }
+            return res.status(200).json({ cancelled, sessionId });
+        } catch (error) {
+            return next(error);
         }
-        return res.status(200).json({ cancelled, sessionId: sessionId || 'default' });
-    });
-
-    app.post('/api/sessions/:sessionId/cancel', async (req, res) => {
-        const { sessionId } = req.params;
-        const cancelled = agent.cancel(sessionId);
-        if (!cancelled) {
-            logger.debug(`No in-flight run to cancel for session: ${sessionId}`);
-        }
-        return res.status(200).json({ cancelled, sessionId });
     });
 
     // Synchronous endpoint: await the full AI response and return it in one go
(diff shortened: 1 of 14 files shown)


src/app/api/server.ts
```diff
@@ -183,6 +183,26 @@ export async function initializeApi(agent: DextoAgent, agentCardOverride?: Parti
...
+    });
+
+    app.post('/api/sessions/:sessionId/cancel', async (req, res) => {
+        const { sessionId } = req.params;
+        const cancelled = agent.cancel(sessionId);
+        if (!cancelled) {
+            logger.debug(`No in-flight run to cancel for session: ${sessionId}`);
+        }
+        return res.status(200).json({ cancelled, sessionId });
+    });
```
rahulkarajgikar: pick 1, we don't need both APIs

leaning towards `/api/sessions/:sessionId/cancel`

also add a lightweight zod parse to validate the input

src/core/llm/services/vercel.ts
```diff
@@ -456,12 +468,26 @@ export class VercelLLMService implements ILLMService {
...
+                        recoverable: true,
+                    });
+                } else {
+                    logger.error(`Error in streamText: ${err?.stack ?? err}`, null, 'red');
+                    this.sessionEventBus.emit('llmservice:error', {
+                        error: err,
+                        context: 'streamText',
+                        recoverable: false,
+                    });
+                }
```
rahulkarajgikar: doesn't look like this code ever executes - tried cancelling and never saw these logs

vercel doesn't trigger it's own `onError` callback in case of aborted signal, it triggers `onAbort`: https://ai-sdk.dev/docs/advanced/stopping-streams


saw this:

```
{"level":"debug","message":"WebSocket received message: {\"type\":\"cancel\",\"sessionId\":\"f0ffbcd6-3d73-47ba-b03e-7a5458e108b9\"}","timestamp":"2025-08-31 15:25:01"}
{"level":"info","message":"Processing cancel command from WebSocket for session: f0ffbcd6-3d73-47ba-b03e-7a5458e108b9.","timestamp":"2025-08-31 15:25:01"}
{"level":"error","message":"Error during DextoAgent.run: No output generated. Check the stream for errors.","timestamp":"2025-08-31 15:25:01"}
{"level":"error","message":"Error processing WebSocket message: No output generated. Check the stream for errors.","timestamp":"2025-08-31 15:25:01"}
{"error":{"name":"AI_NoOutputGeneratedError"},"level":"error","message":"Unhandled WebSocket error: No output generated. Check the stream for errors.","stack":"AI_NoOutputGeneratedError: No output generated. Check the stream for errors.\n    at Object.flush (file:///Users/karaj/Projects/dexto/node_modules/ai/dist/index.mjs:4603:27)\n    at invokePromiseCallback (node:internal/webstreams/util:181:10)\n    at Object.<anonymous> (node:internal/webstreams/util:186:23)\n    at transformStreamDefaultSinkCloseAlgorithm (node:internal/webstreams/transformstream:613:43)\n    at node:internal/webstreams/transformstream:371:11\n    at writableStreamDefaultControllerProcessClose (node:internal/webstreams/writablestream:1153:28)\n    at writableStreamDefaultControllerAdvanceQueueIfNeeded (node:internal/webstreams/writablestream:1233:5)\n    at writableStreamDefaultControllerClose (node:internal/webstreams/writablestream:1200:3)\n    at writableStreamClose (node:internal/webstreams/writablestream:713:3)\n    at writableStreamDefaultWriterClose (node:internal/webstreams/writablestream:1082:10)","timestamp":"2025-08-31 15:25:01","type":"object"}
^C
```


right now: 

```
[finalText, usage, reasoningText] = await Promise.all([
                response.text,
                response.totalUsage,
                response.reasoningText,
            ]);
```

this code is consistently failing with the error above, it's not able to wait and get response.text properly with `Promise.all `when we abort, and the error is bubbling up all the way into dextoagent.run and then the websocket. this is probably expected behaviour for stopped streams. catch in chat-session is also not catching the error because it expects `AbortError` but this is a different error

despite this erroring out in agent.run, currently this implementation works fine because we aren't storing any of the streamed response chunks anywhere in case it's cancelled. so even though all code below in this function for storing response in context, emitting response etc. gets skipped, it doesn't break, it just pretends the ai message never happened.

but ideally we should be storing whatever partial response we have till then into our context (very useful feature in claude code/cursor) 

this is a bit tricky and will require a new way to process reponses in order to properly handle partial responses with different chunks, storing it properly internally without breaking the context (+ other edge cases of when the response is stopped - stopped mid tool execution, stopped when tool needed approval, etc.) and is probably not needed for an initial implementation of cancellation so i will handle this in a separate PR

for now we can revert this code block in `onError` and add a TODO to handle this in `onAbort` when we handle partial responses properly

src/core/llm/services/types.ts
```diff
@@ -22,7 +22,8 @@ export interface ILLMService {
         textInput: string,
         imageData?: ImageData,
         fileData?: FileData,
-        stream?: boolean
+        stream?: boolean,
+        options?: { signal?: AbortSignal }
```
rahulkarajgikar: keep this `options` mandatory, we are always passing it from chat session

src/core/agent/DextoAgent.ts
```diff
@@ -378,6 +378,30 @@ export class DextoAgent {
...
+        this.ensureStarted();
+        const targetSessionId = sessionId || this.currentDefaultSessionId;
+        // Try to use in-memory default session if applicable
+        if (!sessionId && this.defaultSession && this.defaultSession.id === targetSessionId) {
+            return this.defaultSession.cancel();
+        }
+        // If specific session is requested, attempt to fetch from sessionManager cache
+        const existing = this.sessionManager.peekSession?.(targetSessionId) as
+            | ChatSession
+            | undefined;
```
rahulkarajgikar: unnecessary type cast

src/app/webui/components/hooks/useChat.ts
```diff
@@ -531,6 +547,24 @@ export function useChat(wsUrl: string) {
...
+        if (wsRef.current?.readyState === globalThis.WebSocket.OPEN) {
+            wsRef.current.send(JSON.stringify({ type: 'cancel', sessionId }));
+        } else {
+            // Fallback to REST if needed
+            fetch('/api/cancel', {
+                method: 'POST',
+                headers: { 'Content-Type': 'application/json' },
+                body: JSON.stringify({ sessionId }),
+            }).catch(() => void 0);
+        }
```
rahulkarajgikar: don't fallback. websocket should work

src/core/session/chat-session.ts
```diff
@@ -228,14 +234,36 @@ export class ChatSession {
...
+                (typeof error === 'object' && error !== null && (error as any).aborted === true);
+            if (aborted) {
+                this.eventBus.emit('llmservice:error', {
+                    error: new Error('Run cancelled'),
+                    context: 'user_cancelled',
+                    recoverable: true,
+                });
+                return '';
+            }
+            throw error;
```
rahulkarajgikar: todo: currently this only applies for openai, anthropic services, because vercel works differently

we should remove this error handling when we handle partial responses properly in all services

src/core/session/chat-session.ts
```diff
@@ -228,14 +234,36 @@ export class ChatSession {
...
+            // If this was an intentional cancellation, emit a recoverable error event and return empty string
+            const aborted =
+                (error instanceof Error && error.name === 'AbortError') ||
+                (typeof error === 'object' && error !== null && (error as any).aborted === true);
+            if (aborted) {
+                this.eventBus.emit('llmservice:error', {
+                    error: new Error('Run cancelled'),
+                    context: 'user_cancelled',
+                    recoverable: true,
+                });
```
rahulkarajgikar, line 262: todo: remove emit errors, cancellation is a normal state and not an error state. llmservice.completeTask should return the partial computed response till then and not error out in case the execution was cancelled

for now this makes sense but for partial responses we would need to handle this

src/core/session/session-manager.ts
```diff
@@ -293,6 +293,14 @@ export class SessionManager {
         return undefined;
     }
 
+    /**
+     * Returns an in-memory session without loading from storage.
+     * Useful for fast checks (e.g., cancellation) when we do not want to create or restore a session.
+     */
+    public peekSession(sessionId: string): ChatSession | undefined {
+        return this.sessions.get(sessionId);
+    }
```
rahulkarajgikar: add an argument `restoreFromStorage?: boolean` into existing `getSession` and handle this there instead of making this new method - allows any use case to get a session and decide if they want to pull from memory only or storage as well

src/app/cli/cli.ts
```diff
@@ -193,12 +193,47 @@ export async function startAiCli(agent: DextoAgent) {
...
-                        `Error in processing input: ${error instanceof Error ? error.message : String(error)}`
-                    );
+                    const err = error instanceof Error ? error : new Error(String(error));
+                    const aborted =
+                        err.name === 'AbortError' ||
+                        (err as any).aborted === true ||
+                        /abort/i.test(err.message || '');
+                    if (!aborted) {
+                        logger.error(`Error in processing input: ${err.message}`);
+                    }
```
rahulkarajgikar: todo: revert this when we handle partials properly

rahulkarajgikar (commented):
looks good! tested out web ui for some basic cases and works 

we still need to handle partial responses properly and not treat cancellation like an error case + handle some tool edge cases, but i will do that in the next PR along with adding proper integ tests for this

few minor changes + add some todo markers then we are good to merge

shaunak99:
Thanks for the review! The `llmService:error` felt wrong to me too. I've added the TODOs and reverted the vercel block.

src/app/webui/components/hooks/useChat.ts
```diff
@@ -531,6 +547,17 @@ export function useChat(wsUrl: string) {
...
+
+    const cancel = useCallback((sessionId?: string) => {
+        if (wsRef.current?.readyState === globalThis.WebSocket.OPEN) {
+            wsRef.current.send(JSON.stringify({ type: 'cancel', sessionId }));
+        }
+        // Optimistically clear processing state; server will also send events
+        setProcessing(false);
+        // Ensure any ensuing provider stream error from abort does not surface as banner
+        suppressNextErrorRef.current = true;
     }, []);
```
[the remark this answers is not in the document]
rahulkarajgikar: will handle this in #307 where i fixed this for normal websocket messages

src/core/llm/services/vercel.ts
```diff
@@ -186,13 +186,16 @@ export class VercelLLMService implements ILLMService {
 
     async completeTask(
         textInput: string,
+        options: { signal?: AbortSignal },
         imageData?: ImageData,
         fileData?: FileData,
         stream?: boolean
     ): Promise<string> {
```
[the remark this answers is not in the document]
rahulkarajgikar: it's aligned

PR #341, opened 2025-09-11
Revert "chore: version packages"

rahulkarajgikar:
Reverts truffle-ai/dexto#339

## Summary by CodeRabbit

* Chores
  * Synchronized package versions and release metadata to align with repository reorganization.
  * Added change tracking for upcoming patch releases and cleaned up outdated changelog entries.
  * No functional changes to the app, CLI, or APIs; behavior remains unchanged.
  * Maintenance-only update—end users should not notice any differences, and installs/upgrades remain compatible.

### DIFF
--- a/.changeset/stupid-lions-grab.md
+++ b/.changeset/stupid-lions-grab.md
@@ -1,0 +1,7 @@
+---
+'@dexto/webui': patch
+'@dexto/core': patch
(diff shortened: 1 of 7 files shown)


rahulkarajgikar:
temp reverting because the publish failed

Issue #304, opened 2025-08-28
Add client SDK that wraps around APIs


rahulkarajgikar:
We would then use the client SDK instead of directly calling APIs in the web UI/other consumer apps

This allows us to let the client SDK handle the types etc. and improves DX, generally followed practice

shaunak99:
Working on this,  assigning this to myself

rahulkarajgikar:
awesome

rahulkarajgikar:
we need proper input/output zod types for each API as part of this, then the SDK could be just a light wrapper on top

PR #345, opened 2025-09-11
Adding a Client SDK

shaunak99:
This PR introduce a new ClientSDK which is a thin wrapper around our REST APIs with extensive validation logic. This addresses https://github.com/truffle-ai/dexto/issues/304

## Summary by CodeRabbit

* **New Features**
  * Added a lightweight, zero-dependency client SDK: HTTP + optional WebSocket, streaming & non-streaming messaging, sessions, search, LLM/catalog access, MCP tool integration, real-time events, and typed public APIs.

* **Documentation**
  * Added client SDK README with Quick Start and configuration examples.

* **Tests**
  * Added unit tests for client errors, HTTP client behavior, and schema utilities.

* **Refactor**
  * Adjusted browser-facing exports to surface client-oriented types.

* **Chores**
  * Included client SDK in top-level build, added package publish config, and updated lint config to recognize browser fetch/Web APIs.


packages/client-sdk/src/types.ts
```diff
@@ -0,0 +1,241 @@
+// Client SDK types where we intentionally keep API-facing shapes flexible.
+// Wherever possible, prefer importing types from @dexto/core to avoid drift.
+// The types below are kept relaxed (string-based, optional fields) because
+// they mirror HTTP responses and inputs exposed by the public API.
+// We re-export core types in index.ts for strong typing when needed.
+// Import core types for re-export
+import type { InternalMessage } from '@dexto/core';
```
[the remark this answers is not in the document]
shaunak99: @coderabbitai We import from the core package which handles the entry point here so this guideline does not apply

shaunak99:
Have a look if this is good start with. Let me know how you want to version this and if we should call this `@dexto/sdk` instead.

packages/client-sdk/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@dexto/client-sdk",
+  "version": "1.1.4",
```
rahulkarajgikar, line 3: also need to update `.changeset/config.json` which configures how the package versions are managed and synced across each other, to add this new package

(for now we just add it to fixed list with the other 2 packages)

packages/client-sdk/src/client.ts
```diff
@@ -0,0 +1,633 @@
...
+        // Validate inputs with comprehensive Zod validation
+        const validatedConfig = validateInput(ClientConfigSchema, config, 'ClientConfig');
+        const validatedOptions = validateInput(ClientOptionsSchema, options, 'ClientOptions');
+
+        // Apply defaults while avoiding undefined properties for exactOptionalPropertyTypes
+        this.config = {
+            baseUrl: validatedConfig.baseUrl,
+            ...(validatedConfig.apiKey !== undefined ? { apiKey: validatedConfig.apiKey } : {}),
+            timeout: validatedConfig.timeout ?? 30000,
+            retries: validatedConfig.retries ?? 3,
```
rahulkarajgikar: set defaults in zod schema so it's easier to maintain  (same for any other places)

then we can also just directly use `this.config = validatedConfig` without worrying about defaults

(this gets rid of ?? <default> in a lot of places below)

packages/client-sdk/src/client.ts
```diff
@@ -0,0 +1,633 @@
...
+        };
+
+        this.options = {
+            enableWebSocket: validatedOptions?.enableWebSocket ?? true,
+            reconnect: validatedOptions?.reconnect ?? true,
+            reconnectInterval: validatedOptions?.reconnectInterval ?? 5000,
+            debug: validatedOptions?.debug ?? false,
+        };
```
rahulkarajgikar: same here

packages/client-sdk/src/client.ts
```diff
@@ -0,0 +1,633 @@
...
+            );
+        }
+
+        // Connect WebSocket if enabled
+        if (this.options.enableWebSocket) {
+            if (!this.ws) {
+                this.initializeWebSocket();
+            }
+            try {
+                await this.ws!.connect();
```
rahulkarajgikar: avoid `!` which masks undefined issues

packages/client-sdk/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+    test: {
+        environment: 'node',
+        globals: true,
+    },
+});
```
rahulkarajgikar: i think we can delete this file, root vitest.config.ts should work

packages/client-sdk/tsup.config.ts
```diff
@@ -0,0 +1,17 @@
...
+            index: 'src/index.ts',
+        },
+        format: ['cjs', 'esm'],
+        outDir: 'dist',
+        // Generate declaration files without resolving workspace deps
+        dts: { resolve: false },
+        shims: true,
+        bundle: true,
+        platform: 'neutral',
+        external: ['@dexto/core'],
```
rahulkarajgikar: @dexto/core shouldn't be external

external means we don't bundle @dexto/core along with this package build.

this implies that consumers of @dexto/client-sdk will also have to explicitly install @dexto/core to use the client-sdk

it's better to just bundle core along with the client-sdk — remove this line

packages/client-sdk/tests/errors.test.ts
```diff
@@ -0,0 +1,222 @@
+import { describe, it, expect } from 'vitest';
```
rahulkarajgikar: avoid `tests` folder and prefer to keep test files in same folder as code for consistency across the repo

packages/client-sdk/src/client.ts
```diff
@@ -0,0 +1,633 @@
...
+    validateInput,
+    validateResponse,
+} from './schemas.js';
+
+/**
+ * Helper function to safely encode path segments for URLs
+ */
+function encodePathSegment(segment: string): string {
+    return encodeURIComponent(segment);
+}
```
rahulkarajgikar: unnecessary 1 line wrapper, just directly call `encodeURIComponent`

packages/client-sdk/src/schemas.ts
```diff
@@ -0,0 +1,457 @@
...
+
+export const GreetingResponseSchema = z
+    .object({
+        greeting: z.string().nullable(),
+    })
+    .strict();
+
+// ============= UTILITY FUNCTIONS =============
+
+export function validateInput<T>(schema: z.ZodSchema<T>, data: unknown, _context: string): T {
```
rahulkarajgikar: drop the unused `_context`, not necessary, same in `validateResponse`

packages/client-sdk/src/types.ts
```diff
@@ -0,0 +1,225 @@
...
+    response: string;
+    sessionId: string;
+}
+
+export interface SessionInfo {
+    id: string;
+    createdAt: number;
+    lastActivity: number;
+    messageCount: number;
+}
```
rahulkarajgikar: these types are a problem - lot of duplication of the same input shapes we already have in schemas

we now have 3 places with the same types/shapes (core where the struct is defined, then 2 duplicates (zod and type) with the same shape) - will be really annoying to maintain this and will drift super easily with minor changes. 

these should all be derived from 1 source of truth - which should be core

at the very least we could use either z.input or z.output as required to derive these types in client-sdk to just have 1 duplicate here.

haven't checked each type individually but it is likely almost this entire file

ideally we should be figuring out a way to derive everything from the core types to avoid this, so that if core updates, these types also update.

eg:

`SessionInfo` is a known type from core which defines the exact shape

SessionInfoSchema now duplicates the shape once in zod
```
export const SessionInfoSchema = z
    .object({
        id: z.string(),
        createdAt: z.number(),
        lastActivity: z.number(),
        messageCount: z.number().min(0),
    })
    .strict();
```

this interface `SessionInfo` here is a complete copy of the core interface `SessionMetadata`

instead we should follow this approach:

1. if type/shape does not exist in core, add it into core
2. if zod schema is necessary, we can define the zod schema in core itself and derive the necessary types from there, import them here. basically keeping 1 source of truth

rahulkarajgikar: update: looked into this more, so there is a way better solution if we move off `express` and start using `hono` for the API layer instead

Hono is a lightweight express replacement that solves all of these problems and some more

- handles input/output validation of apis better, integrates well with zod
- comes with a fully typed RPC client out of the box
- smaller bundle (12KB vs express 200KB+)
- better edge runtime compatibility (express only works in node-js runtimes, but hono works in any runtime - nodejs/bun/deno/vercel edge runtime/cloudflare workers) with better serverless support and less cold start times -> could be very helpful for deployment flexibility. 

downsides of hono:
- worse middleware support because it's newer (need to write more manual middleware for cors, auth)
- less deep node-js integration
- stuff like static file serving is better in express (we aren't using this)

i think this is probably worth it because we aren't really using these features in express and don't have heavy node-js dependencies anyway

looks like hono seems to be an increasingly popular api framework these days with many ts based projects preferring it over express

with hono we would solve a lot of these problems and the client sdk will become a lot smaller (potentially going from ~1500 lines to about 250-300 lines because we can directly leverage hono's typed client and either directly export that or have lightweight wrapper funcs on it)

rahulkarajgikar: for now:

1. i think we should keep input validation and try to limit duplication here as much as possible by importing these types from core. For apis that don't already have proper input schemas - we should be ideally be adding the types near the api layer. for now since our api layer is tightly coupled to cli, this is a bit difficult though, so we can put these types in core instead and import them here(ideally api server should be its own lib + all the input types defined near the api definitions, and imported here). dependency chain should look like: client-sdk lib ->api lib -> core lib.
 
2. we can completely get rid of output validation of all api responses in the client.ts, it doesn't add a lot of value and is the source of most of the duplication (we already know what api responses will look like) + delete all the dup types related to it. we already know what the underlying response types would be from core - so we can just import those types here and cast the responses to those types (not ideal but better than duplicating)

while i check more about how hono would work

rahulkarajgikar: discussion of what hono code + client-sdk would look like vs current express + client-sdk:

(note how hono auto-types the apis and links well with zod)

```
⏺ ★ Insight ─────────────────────────────────────
  Hono is designed as a drop-in Express replacement with similar middleware patterns but adds powerful features like RPC
  client generation, built-in Zod validation, and type inference. It's particularly popular for its tiny bundle size and
  Edge runtime compatibility.
  ─────────────────────────────────────────────────

  Let me show you how Hono would transform both the Express API and client SDK:

  1. Impact on Express API Code

  Current Express Code:

  // packages/cli/src/api/server.ts
  app.get('/api/sessions', async (req, res, next) => {
      try {
          const sessionIds = await agent.listSessions();
          const sessions = await Promise.all(
              sessionIds.map(async (id) => {
                  const metadata = await agent.getSessionMetadata(id);
                  return {
                      id,
                      createdAt: metadata?.createdAt || null,
                      lastActivity: metadata?.lastActivity || null,
                      messageCount: metadata?.messageCount || 0,
                  };
              })
          );
          return res.json({ sessions });
      } catch (error) {
          return next(error);
      }
  });

  app.post('/api/llm/switch', express.json(), async (req, res, next) => {
      try {
          const body = (req.body ?? {}) as Record<string, unknown>;
          const sessionId = typeof body.sessionId === 'string' ? body.sessionId : undefined;
          const { sessionId: _omit, ...llmCandidate } = body;
          const llmConfig = LLMUpdatesSchema.parse(llmCandidate);
          const config = await agent.switchLLM(llmConfig, sessionId);
          return res.status(200).json({ config, sessionId });
      } catch (error) {
          return next(error);
      }
  });

  Becomes Hono Code:

  // packages/cli/src/api/routes/sessions.ts
  import { Hono } from 'hono';
  import { zValidator } from '@hono/zod-validator';

  // Define schemas once - these become the contract
  const SessionInfoSchema = z.object({
      id: z.string(),
      createdAt: z.number().nullable(),
      lastActivity: z.number().nullable(),
      messageCount: z.number(),
  });

  const SessionsResponseSchema = z.object({
      sessions: z.array(SessionInfoSchema),
  });

  const LLMSwitchRequestSchema = z.object({
      provider: z.string().optional(),
      model: z.string().optional(),
      sessionId: z.string().optional(),
      // ... other fields
  });

  const LLMSwitchResponseSchema = z.object({
      config: LLMConfigSchema,
      sessionId: z.string().optional(),
  });

  // Create typed route handlers
  export const sessionsRoute = new Hono()
      .get('/',
          // Response validation (optional but recommended)
          zValidator('response', SessionsResponseSchema),
          async (c) => {
              const sessionIds = await agent.listSessions();
              const sessions = await Promise.all(
                  sessionIds.map(async (id) => {
                      const metadata = await agent.getSessionMetadata(id);
                      return {
                          id,
                          createdAt: metadata?.createdAt || null,
                          lastActivity: metadata?.lastActivity || null,
                          messageCount: metadata?.messageCount || 0,
                      };
                  })
              );
              return c.json({ sessions }); // Type-checked!
          }
      )
      .post('/:sessionId/load', async (c) => {
          const sessionId = c.req.param('sessionId');
          await agent.loadSession(sessionId);
          return c.json({ status: 'loaded', sessionId });
      });

  export const llmRoute = new Hono()
      .post('/switch',
          zValidator('json', LLMSwitchRequestSchema),
          zValidator('response', LLMSwitchResponseSchema),
          async (c) => {
              const { sessionId, ...llmConfig } = c.req.valid('json');
              const config = await agent.switchLLM(llmConfig, sessionId);
              return c.json({ config, sessionId }, 200);
          }
      );

  // Main app composition
  const app = new Hono()
      .route('/api/sessions', sessionsRoute)
      .route('/api/llm', llmRoute)
      // Error handling is built-in
      .onError((err, c) => {
          if (err instanceof DextoValidationError) {
              return c.json({ error: err.message }, 400);
          }
          return c.json({ error: 'Internal error' }, 500);
      });

  // Export the type for client generation
  export type AppType = typeof app;

  ★ Insight ─────────────────────────────────────
  Notice how Hono's zValidator middleware both validates AND types the request/response. The c.req.valid('json') returns
  a fully typed object - no manual parsing or casting needed. This eliminates the parseBody/parseQuery helpers entirely.
  ─────────────────────────────────────────────────

  Key Changes:

  - Cleaner validation - zValidator middleware instead of manual parsing
  - Type inference - c.req.valid() returns typed data
  - Built-in error handling - .onError() instead of Express error middleware
  - Route composition - .route() for modular organization
  - WebSocket support - Hono has built-in WebSocket handling

  2. Impact on Client SDK Code

  Current Client SDK (633 lines):

  // packages/client-sdk/src/client.ts
  export class DextoClient {
      private http: HttpClient;
      private ws: WebSocketClient | null = null;

      async listSessions(): Promise<SessionInfo[]> {
          const response = await this.http.get<{ sessions: SessionInfo[] }>('/api/sessions');
          const validatedResponse = validateResponse(
              SessionsListResponseSchema,
              response,
              'listSessions'
          );
          return validatedResponse.sessions;
      }

      async switchLLM(config: Partial<LLMConfig>, sessionId?: string): Promise<LLMConfig> {
          const validatedConfig = validateInput(LLMConfigInputSchema, config, 'LLM config');
          if (sessionId !== undefined) {
              validateInput(z.string().min(1), sessionId, 'sessionId');
          }
          const requestBody = {
              ...validatedConfig,
              ...(sessionId && { sessionId }),
          };
          const response = await this.http.post<{ config: LLMConfig }>('/api/llm/switch', requestBody);
          const validatedResponse = validateResponse(LLMSwitchResponseSchema, response, 'switchLLM');
          return validatedResponse.config;
      }
      // ... 500+ more lines
  }

  Becomes Hono RPC Client (~50 lines):

  // packages/client-sdk/src/index.ts
  import { hc } from 'hono/client';
  import type { AppType } from '@dexto/cli/api';

  export class DextoClient {
      private client: ReturnType<typeof hc<AppType>>;
      private ws: WebSocket | null = null;

      constructor(config: { baseUrl: string; apiKey?: string }) {
          this.client = hc<AppType>(config.baseUrl, {
              headers: config.apiKey ? { Authorization: `Bearer ${config.apiKey}` } : {},
          });
      }

      async listSessions() {
          const res = await this.client.api.sessions.$get();
          if (!res.ok) throw new Error(`HTTP ${res.status}`);
          const data = await res.json(); // Fully typed!
          return data.sessions;
      }

      async switchLLM(config: any, sessionId?: string) {
          const res = await this.client.api.llm.switch.$post({
              json: { ...config, sessionId }
          });
          if (!res.ok) throw new Error(`HTTP ${res.status}`);
          const data = await res.json(); // Fully typed!
          return data.config;
      }

      // WebSocket handling remains similar
      connectWebSocket(url: string) {
          this.ws = new WebSocket(url);
          // ... WebSocket event handling
      }
  }

  // That's it! All types are inferred from the server

  What Happens to the SDK Package:

  Deleted files:
  - schemas.ts (435 lines) - ❌ Gone, defined on server
  - types.ts (200+ lines) - ❌ Gone, inferred from server
  - errors.ts - ✅ Keep (still need client-specific error handling)
  - http-client.ts - ❌ Gone, Hono client handles this
  - websocket-client.ts - ✅ Keep (WebSocket logic stays)

  The entire SDK becomes:
  1. A thin wrapper around Hono's RPC client
  2. WebSocket connection management
  3. Client-specific error handling
 ```

maybe we don't even need the thin wrapper and can directly use hono's rpc client, we can see this

Comment (rahulkarajgikar) [reply packages/client-sdk/src/types.ts]: after dropping output validation:

```
 // Before:
  async listSessions(): Promise<SessionInfo[]> {
      const response = await this.http.get<{ sessions: SessionInfo[] }>('/api/sessions');
      const validatedResponse = validateResponse(SessionsListResponseSchema, response, 'listSessions');
      return validatedResponse.sessions;
  }

  // After:
  async listSessions(): Promise<SessionInfo[]> {
      const response = await this.http.get<{ sessions: SessionInfo[] }>('/api/sessions');
      return response.sessions;
  }
```
+ delete the now unused types

(eliminates the need for all the response schemas trimming duplication considerably)

File: packages/client-sdk/src/websocket-client.ts [hunk 12]
```diff
@@ -0,0 +1,282 @@
...
+                this.connect().catch(() => {
+                    // Connection failed, will try again or give up based on max attempts
+                });
+            }
+        }, this.reconnectInterval);
+    }
+
+    private handleIncomingMessage(data: unknown) {
+        const msgData = data as Record<string, unknown>;
+        const event: DextoEvent = {
```
rahulkarajgikar, line 164: use event type from core

packages/client-sdk/src/errors.ts
```diff
@@ -0,0 +1,165 @@
...
+    DextoValidationError,
+    ErrorScope,
+    ErrorType,
+    Issue,
+    type DextoErrorCode,
+} from '@dexto/core';
+
+/**
+ * Client SDK error factory
+ * Creates properly typed errors following core error patterns
```
rahulkarajgikar: this is fine for now but may be unnecessary

the whole idea of these errors was to be only in core so that API middleware could catch any errors from core and transform them into meaningful api responses with correct status codes automatically.

these client errors are on top of the api layer so having these typed errors here doesn't add much value (we do see that we have to cast the error codes to make them fit DextoErrorCode)

but there is no harm as such, it is more informative than regular errors 

can you add a TODO here to re-evaluate these errors - i'm not sure right now what the best thing to do here is but this is fine for now

packages/client-sdk/src/http-client.ts
```diff
@@ -0,0 +1,225 @@
+import { ClientConfig } from './types.js';
+import { ClientError, DextoRuntimeError } from './errors.js';
+
+// Derive fetch types to avoid relying on DOM lib globals in ESLint
+type FetchInput = Parameters<typeof fetch>[0];
+type FetchInit = NonNullable<Parameters<typeof fetch>[1]>;
+type FetchResponse = Awaited<ReturnType<typeof fetch>>;
```
rahulkarajgikar, line 7: what does this mean?

shaunak99: There was some error that was coming up so it had applied this fix, I thought to keep it since you had mentioned a workaround that you had to do in core to support some browser utilities. This is a small brief that I got but let me know if should remove it:

```
⏺ That comment in http-client.ts is about avoiding DOM type pollution in Node.js environments. Here's what it means:

  The Problem

  When TypeScript sees fetch, it could use types from:
  1. DOM lib (lib.dom.d.ts) - browser fetch with Request/Response classes
  2. Node.js (@types/node) - Node's built-in fetch (18+) or undici

  ESLint + TypeScript configs sometimes force DOM types even in Node environments, causing:
  - Type conflicts between browser/Node fetch APIs
  - Unnecessary DOM dependencies in server code
  - Stricter types than needed for generic HTTP client

  The Solution

  // Instead of relying on global fetch types (which might be DOM):
  type FetchInput = Parameters<typeof fetch>[0];   // Whatever fetch() expects as arg 1
  type FetchInit = NonNullable<Parameters<typeof fetch>[1]>; // arg 2, but not undefined  
  type FetchResponse = Awaited<ReturnType<typeof fetch>>;     // What fetch() returns

  This derives types from the actual fetch function available at runtime, rather than importing potentially wrong global types.

  It's actually a clever technique - works whether you're in Node, browser, or using a fetch polyfill.
  ```

packages/client-sdk/tsup.config.ts
shaunak99: I removed this but isn't this the expected behaviour for the client SDK? 

It should not import core but just be a thin wrapper around the APIs with relevant types. Importing node specific packages would also cause issue if we use the SDK in environments where node is not supported (like mobile)

shaunak99: Just a quick analysis overview:

> 
> ## Analysis: Node.js Dependencies in Client SDK
> 
> After thoroughly analyzing your client SDK codebase, I've identified several **critical issues** that will cause problems in mobile and browser environments. Here's my comprehensive analysis:
> 
> ### 🚨 **Critical Issues Found**
> 
> #### 1. **Missing Browser-Safe Type Exports**
> The client SDK imports several types from `@dexto/core` that are **NOT available** in the browser-safe exports:
> 
> **Missing from browser exports:**
> - `SessionMetadata` (from session manager)
> - `SearchOptions`, `SearchResult`, `SessionSearchResult` (from search types)
> - `AgentEventMap`, `SessionEventMap` (from events)
> - `ModelInfo`, `SupportedFileType`, `ProviderInfo` (from LLM types)
> 
> **Current browser exports only include:**
> - `LLMProvider`, `LLMRouter` (enums)
> - `InternalMessage`, `TextPart`, `ImagePart`, `FilePart` (context types)
> - Error types and utilities
> 
> #### 2. **Node.js-Specific Dependencies in Core Package**
> The `@dexto/core` package has **heavy Node.js dependencies** that will cause bundling issues:
> 
> **Problematic dependencies:**
> - `better-sqlite3` - SQLite database (Node.js only)
> - `pg` - PostgreSQL driver (Node.js only)  
> - `ioredis` - Redis client (Node.js only)
> - `winston` - Logging (Node.js only)
> - `dotenv` - Environment variables (Node.js only)
> - `boxen`, `chalk` - Terminal styling (Node.js only)
> 
> **Node.js built-in modules used:**
> - `fs`, `path`, `os`, `crypto`, `events` - Used extensively throughout core
> 
> #### 3. **WebSocket Implementation Issues**
> The WebSocket client has a **hardcoded limitation**:
> 
> ```typescript
> // In websocket-client.ts line 46-57
> if (typeof WebSocket !== 'undefined') {
>     // Browser environment
>     this.ws = new WebSocket(this.url);
> } else {
>     // Node.js environment - would need to import ws package
>     // For now, throw error as Node.js users should use HTTP methods
>     reject(ClientError.websocketConnectionFailed(...));
> }
> ```
> 
> This means **WebSocket functionality is completely broken in Node.js environments**.
> 
> ### �� **Recommended Solutions**
> 
> #### **Option 1: Create Browser-Safe Type Exports (Recommended)**
> Update `/packages/core/src/index.browser.ts` to include the missing types:
> 
> ```typescript
> // Add these exports to index.browser.ts
> export type { SessionMetadata } from './session/session-manager.js';
> export type { 
>     SearchOptions, 
>     SearchResult, 
>     SessionSearchResult,
>     SearchResponse,
>     SessionSearchResponse 
> } from './search/types.js';
> export type { AgentEventMap, SessionEventMap } from './events/index.js';
> export type { ModelInfo, SupportedFileType, ProviderInfo } from './llm/types.js';
> ```
> 
> #### **Option 2: Remove Core Dependency (Better Long-term)**
> Since the client SDK should be a "thin wrapper," consider:
> 
> 1. **Duplicate the needed types** directly in the client SDK
> 2. **Remove the `@dexto/core` dependency** entirely
> 3. **Use only browser-safe utilities** like `zod` for validation
> 
> #### **Option 3: Create Separate Browser Package**
> Create a new `@dexto/core-browser` package with only the types and utilities needed by the client SDK.
> 
> ### 🎯 **Immediate Action Required**
> 
> The client SDK **will fail** in mobile/browser environments because:
> 
> 1. **Type imports will fail** - Missing exports cause TypeScript compilation errors
> 2. **Bundle size will be huge** - All Node.js dependencies get pulled in
> 3. **Runtime errors** - Node.js-specific code will throw errors in browsers
>

shaunak99:
So I tested the updates with the mobile app setup. I think we need to re-evaluate how the client-side input validation needs to be setup if we don't want type duplication. Bundling core as-is has a lot of issues and introduces a lot of bugs when using the SDK for app development due to node runtime dependencies.

Either we need to be able to import types-only from core or create a shared types package and consume those types across packages or come up with a different strategy.

For now, I have removed all validation and reduced the implementation of the Client SDK to a simple lightweight wrapper on our express server endpoints. This works for now in the apps and can be used for app development, just without input validation. Once we figure out the express/hono migration and how to setup types correctly, we can see how to re-implement the client-side input validation.

For now there is considerable type duplication but I guess it's worth starting with since this is the only functional, bug free way I am able to get it to work. Can always remove this later on

rahulkarajgikar:
> So I tested the updates with the mobile app setup. I think we need to re-evaluate how the client-side input validation needs to be setup if we don't want type duplication. Bundling core as-is has a lot of issues and introduces a lot of bugs when using the SDK for app development due to node runtime dependencies.
> 
> Either we need to be able to import types-only from core or create a shared types package and consume those types across packages or come up with a different strategy.
> 
> For now, I have removed all validation and reduced the implementation of the Client SDK to a simple lightweight wrapper on our express server endpoints. This works for now in the apps and can be used for app development, just without input validation. Once we figure out the express/hono migration and how to setup types correctly, we can see how to re-implement the client-side input validation.
> 
> For now there is considerable type duplication but I guess it's worth starting with since this is the only functional, bug free way I am able to get it to work. Can always remove this later on

yeah i found the issue

it's because we are using `bundle: true` in our tsup config for core which bundles the entire repo along with dependencies into 1 index.ts file

so any consumers of dexto/core automatically pull all node related dependencies and any other dependencies (other than the ones marked `external` in tsup config

this is apparently not the right thing to do for libraries, using bundle: false is the way - this keeps the relative paths in the repo and then if you import only types from core, tree-shaking by the bundlers will remove dependencies on node related libraries if they are unused

so with bundle: false, it should solve a lot of these node-related issues when importing just types

(imported types are erased by bundlers compiling typescript to js so they shouldn't cause issues if used with `import type {<type>} from @dexto/core` -> the problem comes when we import any functions/constants from @dexto/core because the core surface isn't browser safe or react-native safe. workarounds are to provide separate export surfaces like we've done for browser, maybe a similar one for react-native. long term the right solution is probably to have proper sub-path exports (dexto/core/logger. dexto/core/storage, etc.)

trying this out to see if this solves these issues

rahulkarajgikar:
so it seems bundle: false doesn't solve all the problems

if you are importing only types though:

setting `import type {type} from @dexto/core` should work (explicitly mention its types) without any issues - typescript removes types from the build if they are explicitly declared

PR #350, opened 2025-09-12
Adding a new PromptsManager

shaunak99:
https://github.com/user-attachments/assets/ea49593f-987c-4f56-b557-b379a30e51bd

## Summary by CodeRabbit

- **New Features**
  - Unified prompts system with configurable starter prompts surfaced as quick-action buttons.

- **Web UI**
  - Welcome/chat show dynamic starter prompts; input supports slash-command autocomplete overlay and improved overflow for overlays.

- **API**
  - New endpoints to list, view, and resolve prompts (with validation and clear errors).

- **CLI**
  - Commands to list and invoke prompts; dynamic per-prompt commands available.

- **Documentation**
  - Added user-facing prompt guides: Code Review, Debug, Explain.

### DIFF
--- a/packages/cli/src/api/server.ts
+++ b/packages/cli/src/api/server.ts
@@ -223,9 +223,12 @@ export async function initializeApi(
         try {
             const name = req.params.name;
             if (!name) throw PromptError.nameRequired();
+            if (!agent.promptsManager) {
+                return sendJsonResponse(res, { error: 'Prompts system unavailable' }, 503);
+            }
             const definition = await agent.promptsManager.getPromptDefinition(name);
             if (!definition) throw PromptError.notFound(name);
-            return res.status(200).json({ definition });
+            return sendJsonResponse(res, { definition }, 200);
         } catch (error) {
             return next(error);
         }
@@ -269,38 +272,14 @@ export async function initializeApi(
             const pr = await agent.promptsManager.getPrompt(resolvedName, args);
             const text = extractPromptText(pr);
             if (!text) throw PromptError.emptyResolvedContent(resolvedName);
-            return res.status(200).json({ text });
+            return sendJsonResponse(res, { text }, 200);
         } catch (error) {
             return next(error);
         }
     });
 
-    // Execute a prompt and send to agent
-    app.post('/api/prompts/:name/execute', express.json(), async (req, res, next) => {
-        try {
-            const name = req.params.name;
-            const schema = z.object({
-                args: z.record(z.string(), z.unknown()).optional(),
-                sessionId: z.string().optional(),
-                stream: z.boolean().optional(),
-            });
-            const body = schema.parse(req.body ?? {});
-            if (!name) throw PromptError.nameRequired();
-            const pr = await agent.promptsManager.getPrompt(name, body.args ?? {});
-            const text = extractPromptText(pr);
-            if (!text) throw PromptError.emptyResolvedContent(name);
-            const response = await agent.run(
-                text,
-                undefined,
-                undefined,
-                body.sessionId,
-                body.stream ?? false
-            );
-            return res.status(200).json({ ok: true, response });
-        } catch (error) {
-            return next(error);
-        }
-    });
+    // Note: We intentionally omit an "execute" endpoint; clients resolve prompts
+    // and then call the regular message endpoint, keeping server surface minimal.
 
     app.post('/api/message', express.json(), async (req, res, next) => {
         logger.info('Received message via POST /api/message');
--- a/packages/webui/components/InputArea.tsx
+++ b/packages/webui/components/InputArea.tsx
@@ -83,7 +83,7 @@ export default function InputArea({ onSend, isSending, variant = 'chat' }: Input
   const [showSlashCommands, setShowSlashCommands] = useState(false);
   const [selectedPrompt, setSelectedPrompt] = useState<{
     name: string;
-    arguments: Array<{ name: string; required: boolean }>;
+    arguments: Array<{ name: string; required?: boolean }>;
   } | null>(null);
 
   const showUserError = (message: string) => {
@@ -156,19 +156,27 @@ export default function InputArea({ onSend, isSending, variant = 'chat' }: Input
     if (!trimmed && !imageData && !fileData) return;
 
     // If slash command typed, resolve to full prompt content at send time
-    if (trimmed.startsWith('/')) {
+    if (trimmed === '/') {
+      // Normalize lone slash to no-op
+      trimmed = '';
+    } else if (trimmed.startsWith('/')) {
       const parsed = parseSlashInput(trimmed);
       const name = parsed.command;
-      const argText = parsed.argsText;
+      // Preserve original suffix including quotes/spacing (trim only leading space)
+      const originalArgsText = trimmed.slice(1 + name.length).trimStart();
       if (name) {
         try {
           const url = new URL(`/api/prompts/${encodeURIComponent(name)}/resolve`, window.location.origin);
+          if (originalArgsText) url.searchParams.set('_context', originalArgsText);
           const res = await fetch(url.toString());
           if (res.ok) {
             const data = await res.json();
             const txt = (data?.text as string) || '';
             if (txt) {
-              trimmed = argText ? `${txt}\n\n${argText}` : txt;
+              // Append original args only if server didn't interpolate them
+              trimmed = originalArgsText && !txt.includes(originalArgsText)
+                ? `${txt}\n\n${originalArgsText}`
+                : txt;
             }
           }
         } catch {
@@ -209,7 +217,10 @@ export default function InputArea({ onSend, isSending, variant = 'chat' }: Input
   };
 
   // Handle prompt selection
-  const handlePromptSelect = (prompt: any) => {
+  const handlePromptSelect = (prompt: { 
+    name: string; 
+    arguments?: Array<{ name: string; required?: boolean }>
+  }) => {
     setSelectedPrompt({ name: prompt.name, arguments: prompt.arguments || [] });
     const slash = `/${prompt.name}`;
     setText(slash);


packages/core/src/agent/DextoAgent.ts
```diff
@@ -102,8 +103,13 @@ export class DextoAgent {
      * But the main recommended entry points/functions would still be the wrapper methods we define below
      */
     public readonly mcpManager!: MCPManager;
-    public readonly promptManager!: PromptManager;
+    public readonly systemPromptManager!: SystemPromptManager;
+    // Deprecated: backward-compat alias for older integrations
+    public get promptManager(): SystemPromptManager {
+        return this.systemPromptManager;
+    }
```
rahulkarajgikar, line 110: delete this

packages/core/src/agent/schemas.ts
```diff
@@ -114,6 +114,69 @@ export const AgentConfigSchema = z
...
+                        category: z
+                            .enum(['general', 'coding', 'analysis', 'tools', 'learning'])
+                            .optional()
+                            .default('general')
+                            .describe('Category for organizing starter prompts'),
+                        icon: z
+                            .string()
+                            .optional()
+                            .default('')
+                            .describe('Emoji or icon to display'),
```
rahulkarajgikar, line 152: we can drop icon and let title handle this

packages/core/src/utils/service-initializer.ts
```diff
@@ -41,7 +41,9 @@ import { AgentEventBus } from '../events/index.js';
 export type AgentServices = {
     mcpManager: MCPManager;
     toolManager: ToolManager;
-    promptManager: PromptManager;
+    systemPromptManager: SystemPromptManager;
+    // Deprecated: use systemPromptManager instead. Kept for backward compatibility.
```
rahulkarajgikar, line 45: remove, avoid unnecessary backward compat

packages/webui/components/ChatApp.tsx
```diff
@@ -62,6 +62,26 @@ export default function ChatApp() {
   // Welcome screen search state
   const [welcomeSearchQuery, setWelcomeSearchQuery] = useState('');
 
+  // Starter prompts state (from agent config via /api/prompts)
+  const [starterPrompts, setStarterPrompts] = useState<any[]>([]);
```
rahulkarajgikar, line 66: don't useState with `any`, import the right type

packages/core/src/agent/schemas.ts
```diff
@@ -114,6 +114,69 @@ export const AgentConfigSchema = z
...
+                            path: ['starterPrompts', idx, 'id'],
+                        });
+                    } else {
+                        seen.set(p.id, idx);
+                    }
+                });
+            })
+            .transform((arr) =>
+                arr.map((p) => ({ ...p, title: p.title ?? p.id.replace(/-/g, ' ') }))
+            )
```
rahulkarajgikar, line 177: we can make title non-optional and drop this transform

also update any necessary types that are derived from this

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+        this.providers.set('mcp', new MCPPromptProvider(mcpManager));
+        this.providers.set('internal', new InternalPromptProvider(promptsDir));
+        this.providers.set('starter', new StarterPromptProvider(agentConfig));
+
+        logger.debug(
+            `PromptsManager initialized with providers: ${Array.from(this.providers.keys()).join(', ')}`
+        );
+
+        // If an event bus is provided and supports subscriptions, subscribe to MCP-related events
+        if (this.eventBus && typeof (this.eventBus as any).on === 'function') {
```
rahulkarajgikar, line 56: this defensive check is unnecssary, eventBus.on should always be defined

just make eventBus mandatory and we should be able to avoid this if-else tree entirely

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
+import type { MCPManager } from '../mcp/manager.js';
```
rahulkarajgikar, line 1: rename file to manager.ts for consistency

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+    private promptsCache: PromptSet = {};
+    private cacheValid: boolean = false;
+    private aliasMap = new Map<string, { baseName: string; providerName: string }>();
+    private buildPromise: Promise<void> | null = null;
+
+    constructor(
+        mcpManager: MCPManager,
+        promptsDir?: string,
+        agentConfig?: ValidatedAgentConfig,
+        private eventBus?: AgentEventBus
```
rahulkarajgikar, line 44: all of these should be mandatory arguments, they are always known and always defined

avoid using optionals unless it makes sense

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+                        // Store the alias mapping for provider lookups
+                        this.aliasMap.set(name, { baseName, providerName });
+                    }
+                    allPrompts[name] = name === baseName ? p : { ...p, name };
+                });
+
+                totalPromptsCount += providerPrompts.length;
+                logger.debug(
+                    `📝 Cached ${providerPrompts.length} prompts from ${providerName} provider`
+                );
```
rahulkarajgikar, line 167: this conflict resolution is overly complex

we can avoid aliasMap and all of this

just always use the source prefix - this is what we have done for tools


```
Current: Complex conflict resolution with aliasing and
  prefixing
  Simplification: Always use provider-prefixed names
  (mcp:prompt, internal:prompt, starter:prompt)
  // Instead of complex aliasing, just always prefix:
  const promptKey =
  `${provider.getSource()}:${promptName}`;
```

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+        return Object.prototype.hasOwnProperty.call(prompts, name);
+    }
+
+    /**
+     * Check if a prompt exists (alias for has)
+     */
+    async hasPrompt(name: string): Promise<boolean> {
+        return this.has(name);
+    }
+
```
rahulkarajgikar, line 222: unnecessary duplication - has and hasPrompt are the same
keep one if it's used, delete if neither are used

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        const prompts = Object.values(this.promptsCache).map((prompt) =>
+            this.sanitizePromptMetadata(prompt)
+        );
+
+        // For now, return all prompts without pagination
+        // TODO: Implement proper pagination when needed
+        return {
+            prompts,
```
rahulkarajgikar, line 238: this is just a duplicate of `list` - delete

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+                typeof meta?.originalName === 'string' ? (meta!.originalName as string) : undefined;
+            const command =
+                typeof meta?.command === 'string' ? (meta!.command as string) : undefined;
+
+            if (originalName === nameOrAlias) return key;
+            if (command && (command === want || command === wantNoSlash)) return key;
+        }
+
+        return null;
+    }
```
rahulkarajgikar, line 361: using prefixes everywhere might remove the need for this method as well

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+        if (!promptInfo) {
+            return null;
+        }
+
+        return {
+            name: promptInfo.name,
+            ...(promptInfo.title && { title: promptInfo.title }),
+            ...(promptInfo.description && { description: promptInfo.description }),
+            ...(promptInfo.arguments && { arguments: promptInfo.arguments }),
+        };
```
rahulkarajgikar, line 260: promptInfo is already typed as PromptDefinition.. we can just return promptInfo

we don't even need the null check

packages/core/src/prompts/providers/internal-prompt-provider.ts
```diff
@@ -0,0 +1,310 @@
...
+                // Filter for markdown files
+                const markdownFiles = files.filter((file) => extname(file).toLowerCase() === '.md');
+
+                for (const file of markdownFiles) {
+                    try {
+                        const promptName = file.replace(/\.md$/, '');
+                        const filePath = join(this.promptsDir, file);
+
+                        // Read the markdown content
+                        const content = await readFile(filePath, 'utf-8');
```
rahulkarajgikar, line 60: we don't need caching at both layers

we can keep providers stateless and just cache at the manager level

let providers call the entire methods each time, let manager handle caching

applies to other prompt providers as well

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+
+        const searchTerm = query.toLowerCase();
+        const results = Object.values(this.promptsCache).filter(
+            (prompt) =>
+                prompt.name.toLowerCase().includes(searchTerm) ||
+                (prompt.description && prompt.description.toLowerCase().includes(searchTerm)) ||
+                (prompt.title && prompt.title.toLowerCase().includes(searchTerm))
+        );
+        return results.map((p) => this.sanitizePromptMetadata(p));
+    }
```
rahulkarajgikar, line 292: unused code, delete

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+            ...(promptInfo.title && { title: promptInfo.title }),
+            ...(promptInfo.description && { description: promptInfo.description }),
+            ...(promptInfo.arguments && { arguments: promptInfo.arguments }),
+        };
+    }
+
+    /**
+     * Get prompts by source
+     */
+    async getPromptsBySource(source: 'mcp' | 'internal' | 'starter'): Promise<PromptInfo[]> {
```
rahulkarajgikar, line 266: unused, delete

let's not keep unused code

we can always add these functions later if they are necessary

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+            return promptWithoutMetadata;
+        }
+
+        return { ...prompt, metadata: safeMetadata };
+    }
+
+    /**
+     * Update starter prompts configuration (updates the starter provider)
+     */
+    updateStarterPrompts(agentConfig?: ValidatedAgentConfig): void {
```
rahulkarajgikar, line 436: unused, delete

packages/core/src/prompts/providers/internal-prompt-provider.ts
```diff
@@ -0,0 +1,310 @@
...
+ * as prompts. The filename (without extension) becomes the prompt identifier.
+ * It implements the MCP specification for prompt discovery and retrieval.
+ */
+export class InternalPromptProvider implements PromptProvider {
+    private promptsDir: string;
+    private promptsCache: PromptInfo[] = [];
+    private cacheValid: boolean = false;
+
+    constructor(promptsDir: string = 'prompts') {
+        this.promptsDir = promptsDir;
```
rahulkarajgikar, line 21: i think we should move this `prompts` folder with the markdown files to be within the core repo `prompts` folder

and hardcode the location of the directory here 

it's easier to just keep them within the core prompts/providers folder

siimlar to internal tools which are contained within the repo.

for user provided ones, we could have FilePromptProvider and let them provide relative paths to the prompts in their agent configs (@agentdir/debug-prompt.md or something like this - similar to file system prompts)

i.e user defined prompts - colocated with their agents
internal prompts - colocated with internal code

packages/core/src/prompts/providers/internal-prompt-provider.ts
```diff
@@ -0,0 +1,310 @@
...
+                `Failed to get prompt '${name}': ${error instanceof Error ? error.message : String(error)}`
+            );
+            throw error; // Re-throw as GetPromptResult doesn't allow null
+        }
+    }
+
+    /**
+     * List all available internal prompts
+     */
+    async listPrompts(_cursor?: string): Promise<PromptListResult> {
```
rahulkarajgikar, line 249: we don't need this cursor arg -> we can remove from the interface

packages/core/src/prompts/providers/internal-prompt-provider.ts
```diff
@@ -0,0 +1,310 @@
...
+                messages: [
+                    {
+                        role: 'user',
+                        content: {
+                            type: 'text',
+                            text: content.trim(),
+                        },
+                    },
+                ],
+            };
```
rahulkarajgikar, line 237: we can just return the text here right

in the usages of this method, we are re-extracting out the text content anyway.

i don't see why we need this messages format — even if MCP is using this, we don't need to - we can make our interface return text [currently we have conversion from text -> MCP type -> back to text.]

just returning text fits in better with our system (end user doesn't actually see or need any of this messages shape)

so we can update this interface method to return a string, and handle the conversion in the mcp prompt provider to convert from mcp prompt spec to text.

packages/core/src/prompts/providers/internal-prompt-provider.ts
```diff
@@ -0,0 +1,310 @@
...
+            prompts: safePrompts,
+        };
+    }
+
+    /**
+     * Check if a prompt exists
+     */
+    async hasPrompt(name: string): Promise<boolean> {
+        const prompts = await this.listPrompts();
+        return prompts.prompts.some((prompt: PromptInfo) => prompt.name === name);
```
rahulkarajgikar, line 281: we don't need this method

packages/core/src/prompts/providers/internal-prompt-provider.ts
rahulkarajgikar: it might even be worth thinking if internal prompts are required.

i see many downsides which are better handled by making it file based and configuring it via YAML:
 - we need to copy these md files into the build/dist output so that it doesn't break
 - UI will show these prompts regardless of what the actual agent that is running is (debug prompt for an image editing agent would be confusing)
 - users will see prompts that they haven't configured in YAML which is kind of opaque behaviour

instead, we allow file based prompts, user specifies the file in their YAML using @agent_dir.
(similar to file based system prompt contributor, we can re-use that implementation almost completely)

this has advantages:
- we can associate prompts with specific agents only (coding agent gets coding related prompts, image agent gets its own prompts etc. etc.)
- the prompts would also be colocated with agents itself which is better, and UI + agent behaviour is fully customizable through YAML
- users only see what they have customized

packages/core/src/prompts/types.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Invalidate the provider's internal cache
+     */
+    invalidateCache(): void;
+
+    /**
+     * List all available prompts from this provider with pagination support
+     */
+    listPrompts(cursor?: string): Promise<PromptListResult>;
```
rahulkarajgikar, line 127: drop cursor

we don't need pagination support until we hit much higher scale (which i doubt we will for prompt templates)

packages/core/src/prompts/types.ts
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Get a specific prompt by name
+     */
+    getPrompt(name: string, args?: Record<string, unknown>): Promise<GetPromptResult>;
+
+    /**
+     * Check if a prompt exists
+     */
+    hasPrompt(name: string): Promise<boolean>;
+
```
rahulkarajgikar, line 138: don't think we need this method tbh, doesn't seem to be used anywhere

getPrompt might be sufficient

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,443 @@
...
+    /**
+     * Get all registered provider sources
+     */
+    getProviderSources(): string[] {
+        return Array.from(this.providers.keys());
+    }
+
+    /**
+     * Sanitize prompt metadata to remove sensitive information
+     */
```
rahulkarajgikar, line 414: i don't get this function, what is this for?

prompt.metadata is a Record<string, unknown> type

these things we are filtering out from metadata:

```
            content: _content,
            prompt: _promptText,
            filePath: _filePath,
            messages: _messages,
```

is this something specific to MCP prompts? where are these fields coming from?

packages/core/src/prompts/types.ts
```diff
@@ -0,0 +1,143 @@
...
+        text?: string;
+        data?: string; // base64-encoded blob data
+    };
+}
+
+/**
+ * Union type for all content types
+ */
+export type PromptMessageContent = TextContent | ImageContent | AudioContent | ResourceContent;
+
```
rahulkarajgikar, line 80: a lot of these seem like they might be duplicates of the types we use for storing messages internally

currently our internal message types aren't proper though so we can add a TODO to re-check thse types later against the internal message types

PR #354, opened 2025-09-14
Add a new ResourceManager

shaunak99:
- Ported ResourceManager into packages/core, wired into service initializer and DextoAgent, and added REST endpoints to list/read resources.
  - WebUI now supports @-mention resource autocomplete (masked names); server-side expansion resolves and inlines resource content before sending to the model.

## Summary by CodeRabbit

* **New Features**
  * HTTP APIs to list, read and check MCP and internal filesystem resources.
  * Agent exposes runtime resource management and expands in-message @resource references before model calls.
  * Web UI: @-mention resource autocomplete with keyboard navigation and a useResources hook.

* **Improvements**
  * ResourceManager, internal resources provider and filesystem handler with refresh/cache support.
  * Structured resource error codes and typed resource errors; new config schema and consolidated resource types/exports.

### DIFF
--- a/packages/core/src/resources/errors.ts
+++ b/packages/core/src/resources/errors.ts
@@ -7,14 +7,44 @@ import { ResourceErrorCodes } from './error-codes.js';
  * Creates properly typed errors for resource operations
  */
 export class ResourceError {
+    private static redactUri(uri: string): string {
+        try {
+            const u = new URL(uri);
+            if (u.username) u.username = '***';
+            if (u.password) u.password = '***';
+            u.searchParams.forEach((_, k) => {
+                if (/token|key|secret|sig|pwd|password/i.test(k)) u.searchParams.set(k, '***');
+            });
+            return u.toString();
+        } catch {
+            return uri
+                .replace(/\/\/([^@]+)@/, '//***@')
+                .replace(/((?:token|key|secret|sig|pwd|password)=)[^&]*/gi, '$1***');
+        }
+    }
+
+    private static toMessageAndRaw(reason: unknown): { message: string; raw: unknown } {
+        if (reason instanceof Error) {
+            return {
+                message: reason.message,
+                raw: { name: reason.name, message: reason.message, stack: reason.stack },
+            };
+        }
+        if (typeof reason === 'string') return { message: reason, raw: reason };
+        try {
+            return { message: JSON.stringify(reason), raw: reason };
+        } catch {
+            return { message: String(reason), raw: reason };
+        }
+    }
     // URI format and parsing errors
     static invalidUriFormat(uri: string, expected?: string) {
         return new DextoRuntimeError(
             ResourceErrorCodes.INVALID_URI_FORMAT,
             ErrorScope.RESOURCE,
             ErrorType.USER,
-            `Invalid resource URI format: '${uri}'${expected ? ` (expected ${expected})` : ''}`,
-            { uri, expected },
+            `Invalid resource URI format: '${ResourceError.redactUri(uri)}'${expected ? ` (expected ${expected})` : ''}`,
+            { uri: ResourceError.redactUri(uri), uriRaw: uri, expected },
             expected ? `Use format: ${expected}` : 'Check the resource URI format'
         );
     }
@@ -36,8 +66,8 @@ export class ResourceError {
             ResourceErrorCodes.RESOURCE_NOT_FOUND,
             ErrorScope.RESOURCE,
             ErrorType.NOT_FOUND,
-            `Resource not found: '${uri}'`,
-            { uri },
+            `Resource not found: '${ResourceError.redactUri(uri)}'`,
+            { uri: ResourceError.redactUri(uri), uriRaw: uri },
             'Check that the resource exists and is accessible'
         );
     }
@@ -47,8 +77,8 @@ export class ResourceError {
             ResourceErrorCodes.PROVIDER_NOT_INITIALIZED,
             ErrorScope.RESOURCE,
             ErrorType.SYSTEM,
-            `${providerType} resource provider not initialized for: '${uri}'`,
-            { providerType, uri },
+            `${providerType} resource provider not initialized for: '${ResourceError.redactUri(uri)}'`,
+            { providerType, uri: ResourceError.redactUri(uri), uriRaw: uri },
             'Ensure the resource provider is properly configured'
         );
     }
@@ -66,17 +96,13 @@ export class ResourceError {
 
     // Content access errors
     static readFailed(uri: string, reason: unknown) {
-        const reasonMsg = reason instanceof Error ? reason.message : String(reason);
-        const reasonRaw =
-            reason instanceof Error
-                ? { name: reason.name, message: reason.message, stack: reason.stack }
-                : reason;
+        const { message: reasonMsg, raw: reasonRaw } = ResourceError.toMessageAndRaw(reason);
         return new DextoRuntimeError(
             ResourceErrorCodes.READ_FAILED,
             ErrorScope.RESOURCE,
             ErrorType.SYSTEM,
-            `Failed to read resource '${uri}': ${reasonMsg}`,
-            { uri, reason: reasonMsg, reasonRaw },
+            `Failed to read resource '${ResourceError.redactUri(uri)}': ${reasonMsg}`,
+            { uri: ResourceError.redactUri(uri), uriRaw: uri, reason: reasonMsg, reasonRaw },
             'Check resource permissions and availability'
         );
     }
@@ -86,8 +112,8 @@ export class ResourceError {
             ResourceErrorCodes.ACCESS_DENIED,
             ErrorScope.RESOURCE,
             ErrorType.FORBIDDEN,
-            `Access denied to resource: '${uri}'`,
-            { uri },
+            `Access denied to resource: '${ResourceError.redactUri(uri)}'`,
+            { uri: ResourceError.redactUri(uri), uriRaw: uri },
             'Ensure you have permission to access this resource'
         );
     }
@@ -98,18 +124,14 @@ export class ResourceError {
             ResourceErrorCodes.NO_SUITABLE_PROVIDER,
             ErrorScope.RESOURCE,
             ErrorType.NOT_FOUND,
-            `No suitable provider found for resource: '${uri}'`,
-            { uri },
+            `No suitable provider found for resource: '${ResourceError.redactUri(uri)}'`,
+            { uri: ResourceError.redactUri(uri), uriRaw: uri },
             'Check that the resource type is supported'
         );
     }
 
     static providerError(providerType: string, operation: string, reason: unknown) {
-        const reasonMsg = reason instanceof Error ? reason.message : String(reason);
-        const reasonRaw =
-            reason instanceof Error
-                ? { name: reason.name, message: reason.message, stack: reason.stack }
-                : reason;
+        const { message: reasonMsg, raw: reasonRaw } = ResourceError.toMessageAndRaw(reason);
         return new DextoRuntimeError(
             ResourceErrorCodes.PROVIDER_ERROR,
             ErrorScope.RESOURCE,


packages/cli/src/api/server.ts
```diff
@@ -154,6 +155,34 @@ export async function initializeApi(
...
+        return new DextoValidationError([
+            {
+                code: AgentErrorCode.API_VALIDATION_ERROR,
+                message: errorMessage,
+                scope: ErrorScope.AGENT,
+                type: ErrorType.USER,
+                severity: 'error' as const,
+            },
+        ]);
+    }
```
rahulkarajgikar, line 171: middleware already handles zod errors, this is not needed

packages/cli/src/api/server.ts
```diff
@@ -154,6 +155,34 @@ export async function initializeApi(
...
+        } catch (error) {
+            const errorMessage = error instanceof Error ? error.message : 'URI decode error';
+            logger.error(`Failed to decode resourceId parameter: ${errorMessage}`);
+            throw ResourceError.invalidUriFormat(
+                resourceIdParam,
+                'valid URI-encoded resource identifier'
+            );
+        }
+    }
+
```
rahulkarajgikar, line 185: avoid one line wrapper functions

either directly use decodeURIComponent and handle errors in callers or move this error handling logic into decodeURIComponent

packages/cli/src/api/server.ts
```diff
@@ -371,6 +400,132 @@ export async function initializeApi(
...
+                    {
+                        code: AgentErrorCode.API_VALIDATION_ERROR,
+                        message: 'Missing serverId or resourceId parameters',
+                        scope: ErrorScope.AGENT,
+                        type: ErrorType.USER,
+                        severity: 'error' as const,
+                    },
+                ])
+            );
+        }
```
rahulkarajgikar, line 498: use a lightweight zod schema and let middleware handle this instead

packages/cli/src/api/server.ts
```diff
@@ -371,6 +400,132 @@ export async function initializeApi(
...
+        try {
+            const qualifiedUri = `mcp:${serverId}:${decodedResourceId}`;
+            const content = await agent.readResource(qualifiedUri);
+            return sendJsonResponse(res, { success: true, data: { content } }, 200);
+        } catch (error) {
+            const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+            logger.error(
+                `Error reading resource '${decodedResourceId}' from server '${serverId}': ${errorMessage}`
+            );
+            return next(error);
```
rahulkarajgikar, line 525: avoid try catch blocks like this, this is the whole point of the error handling middleware

- core returns typed errors 
- api routes should just route errors to middleware which handles the errors

error handling middelware already handles zod, and Dexto internal errors

this allows a cleaner api surface and eliminates all this boilerplate

update all apis introduced in this pr to follow this pattern

there should just be one catch block which routes to middleware

packages/core/src/resources/error-codes.ts
```diff
@@ -0,0 +1,13 @@
...
+    RESOURCE_NOT_FOUND: 'resource_not_found',
+    PROVIDER_NOT_INITIALIZED: 'resource_provider_not_initialized',
+    PROVIDER_NOT_AVAILABLE: 'resource_provider_not_available',
+    READ_FAILED: 'resource_read_failed',
+    ACCESS_DENIED: 'resource_access_denied',
+    NO_SUITABLE_PROVIDER: 'resource_no_suitable_provider',
+    PROVIDER_ERROR: 'resource_provider_error',
+} as const;
+
+export type ResourceErrorCode = (typeof ResourceErrorCodes)[keyof typeof ResourceErrorCodes];
```
rahulkarajgikar, line 13: use an enum, this is not consistent with the rest of the repo

packages/core/src/resources/core.ts
```diff
@@ -0,0 +1,18 @@
...
+    ResourceProvider,
+    ResourceSet,
+} from './types.js';
+
+export type { InternalResourcesConfig, ValidatedInternalResourcesConfig } from './schemas.js';
+
+export { ResourceManager } from './manager.js';
+
+export { ResourceError } from './errors.js';
+export { ResourceErrorCodes } from './error-codes.js';
```
rahulkarajgikar, line 18: we don't need this file, 

just keep index.ts for exports

packages/core/src/resources/internal.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
rahulkarajgikar, line 1: we don't need this file, let index.ts handle exports

packages/core/src/resources/index.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Resource management module exports
+ * Organized into thematic sub-barrels for better tree-shaking and maintainability
+ */
+
+// Re-export from thematic sub-barrels
+export * from './core.js';
+export * from './internal.js';
+export * from './reference.js';
```
rahulkarajgikar, line 9: too many sub-files just for handling exports, just let index.ts manage this, and delete core.ts, internal.ts, refererence.ts

it doesn't improve maintainability, 1 source of truth is better, especially for small modules with <10 files

and we can solve tree-shaking by using explicit exports anyway (avoiding barrel exports with *, export named funcs, types etc.)

packages/core/src/resources/types.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * Core resource types and interfaces for the ResourceManager
+ */
+
+import type { ReadResourceResult } from '@modelcontextprotocol/sdk/types.js';
+
+/**
+ * Supported resource sources
+ */
+export type ResourceSource = 'mcp' | 'plugin' | 'custom';
```
rahulkarajgikar, line 10: keep this minimal to only what we currently support in this PR (no plugin)

packages/core/src/resources/schemas.ts
```diff
@@ -0,0 +1,116 @@
...
+    FileSystemResourceSchema,
+]);
+
+/**
+ * Schema for internal resources configuration with smart auto-enable logic
+ *
+ * Design principles:
+ * - Clean input format: just specify resources array or object
+ * - Auto-enable when resources are specified
+ * - Backward compatibility with explicit enabled field
```
rahulkarajgikar, line 86: backward compat with what?

packages/core/src/resources/internal-provider.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export class InternalResourcesProvider implements ResourceProvider {
+    private config: ValidatedInternalResourcesConfig;
+    private handlers: Map<string, InternalResourceHandler> = new Map();
+    private services: InternalResourceServices;
+
+    constructor(config: ValidatedInternalResourcesConfig, services: InternalResourceServices = {}) {
+        this.config = config;
+        this.services = services;
+        logger.debug('InternalResourcesProvider initialized with config:', config);
```
rahulkarajgikar, line 21: nit: fix log to use ``

packages/core/src/resources/internal-provider.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        for (const resourceConfig of this.config.resources) {
+            try {
+                const handler = createInternalResourceHandler(resourceConfig.type);
+                const parsedConfig = InternalResourceConfigSchema.parse(resourceConfig);
+                await handler.initialize(parsedConfig, this.services);
+                this.handlers.set(resourceConfig.type, handler);
+                logger.debug(`Initialized ${resourceConfig.type} resource handler`);
+            } catch (error) {
+                logger.error(`Failed to initialize ${resourceConfig.type} resource handler`, error);
```
rahulkarajgikar, line 38: nit: fix logs

check other log usages in this PR as well to avoid commas which won't print the error
(i see many)

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,426 @@
...
+        '.php',
+        '.rb',
+        '.go',
+        '.rs',
+        '.java',
+        '.kt',
+        '.swift',
+        '.sql',
+        '.sh',
+        '.bash',
```
rahulkarajgikar, line 63: this is a duplicate of what we use in schemas

check for other constant/file extension duplications

move them to constants.ts and import them where required

packages/core/src/resources/manager.ts
```diff
@@ -0,0 +1,182 @@
...
+
+export interface ResourceManagerOptions {
+    internalResourcesConfig?: ValidatedInternalResourcesConfig;
+}
+
+export class ResourceManager {
+    private mcpManager: MCPManager;
+    private internalResourcesProvider?: InternalResourcesProvider;
+
+    constructor(mcpManager: MCPManager, options?: ResourceManagerOptions) {
```
rahulkarajgikar, line 17: change this from options to config and remove the ResourceManagerOptions interface which is an  unnecessary layer of indirection for just one argument

packages/core/src/resources/manager.ts
```diff
@@ -0,0 +1,182 @@
...
+        this.mcpManager = mcpManager;
+        if (options?.internalResourcesConfig?.enabled) {
+            this.internalResourcesProvider = new InternalResourcesProvider(
+                options.internalResourcesConfig
+            );
+        }
+        logger.debug('ResourceManager initialized as thin coordinator');
```
rahulkarajgikar, line 24: remove this log, we can rely on the initiialize log

packages/core/src/resources/manager.ts
```diff
@@ -0,0 +1,182 @@
...
+                logger.error(
+                    `Failed to get internal resources: ${error instanceof Error ? error.message : String(error)}`
+                );
+            }
+        }
+
+        const totalResources = Object.keys(resources).length;
+        const mcpCount = Object.keys(resources).filter((k) => k.startsWith('mcp:')).length;
+        const internalCount = Object.keys(resources).filter((k) =>
+            k.startsWith('internal:')
```
rahulkarajgikar, line 106: nit: increment a counter in the above loops while we are adding the resources, instead of filtering through Object.keys each time which adds O(n) for each [O(3n) vs O(n)]

(minor but could become a problem if we have hundreds of resources through directories)

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,426 @@
...
+            '.hpp': 'text/x-c++',
+            '.sh': 'text/x-shellscript',
+            '.bash': 'text/x-shellscript',
+            '.zsh': 'text/x-shellscript',
+            '.fish': 'text/x-shellscript',
+            '.sql': 'text/x-sql',
+            '.rst': 'text/x-rst',
+            '.tex': 'text/x-tex',
+            '.dockerfile': 'text/x-dockerfile',
+        };
```
rahulkarajgikar, line 414: move this to constants.ts, siomilar to supported types

we can maintain one map of supported extensions -> mimetypes

derive the supported extensions from the keys of the map
use supported mimetypes here

reduces overall duplication with a similar approach to LLM registry

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,426 @@
...
+        } catch {
+            return;
+        }
+        if (!this.isPathAllowed(canonical)) return;
+
+        const maxDepth = this.config?.maxDepth ?? FileSystemResourceHandler.DEFAULT_MAX_DEPTH;
+        const maxFiles = this.config?.maxFiles ?? FileSystemResourceHandler.DEFAULT_MAX_FILES;
+        const includeHidden = this.config?.includeHidden ?? false;
+        const includeExtensions =
+            this.config?.includeExtensions ?? FileSystemResourceHandler.DEFAULT_INCLUDE_EXTENSIONS;
```
rahulkarajgikar, line 280: set defaults in config zod, don't use defaults here

by this point these things should be set

with this we can likely get rid of the above duplicate constants

packages/core/src/resources/internal-registry.ts
rahulkarajgikar: update: delete if not reqd (check below)

packages/core/src/resources/reference.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Resource reference parsing and expansion
+ */
+
+export {
+    parseResourceReferences,
+    resolveResourceReferences,
+    expandMessageReferences,
+    formatResourceContent,
+} from './reference-parser.js';
+export type { ResourceReference, ResourceExpansionResult } from './reference-parser.js';
```
rahulkarajgikar, line 11: delete

packages/core/src/resources/internal-provider.ts
```diff
@@ -0,0 +1,157 @@
...
+import {
+    InternalResourceConfig,
+    InternalResourceHandler,
+    InternalResourceServices,
+    createInternalResourceHandler,
+} from './internal-registry.js';
+import type { ValidatedInternalResourcesConfig } from './schemas.js';
+import { InternalResourceConfigSchema } from './schemas.js';
+
+export class InternalResourcesProvider implements ResourceProvider {
```
rahulkarajgikar, line 13: i dont understand why we need an `InternalResourcesProvider` and separate `internalresourcehandler`

it feels like over-engineering to me -> we have 2 layers of abstractions (internalresourceprovider and internalresourcehandler) for handling 1 use case of files 

do we anticipate any other types beyond `file`/`filesystem` for internal resources? 
also what exactly are we going for with the name `internal` here -> does this imply the resources will be stored? or internal to the agent? 

in general it's better to start with a simpler system that can evolve in terms of complexity/abstraction as requirements get added (i've been guilty of overcomplicating some of our previous implementations as well and ultimately refactored to simpler implementations when i realized the abstractions i assumed we would need were bad)

initial thought: could we just have a top level resource manager and a FileResourceProvider?

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,426 @@
+import { ResourceMetadata } from './types.js';
```
rahulkarajgikar, line 1: this file name is confusing

if we do decide to continue with the `internal` naming, then:

make a separate folder for internal resource management: `resources/internal`

to this folder:

- add types.ts which defines the interface shapes/types (currently all handled in internal-registry.ts)

- add filehandler.ts which defines the `FileSystemResourceHandler`

- add `provider.ts` that aggregates the different handlers (current internal-provider.ts) 

this is also more like our tools modules which follows the above structure for internal tools

this better separates the layers -> manager.ts is top level aggregator
internal -> handles internal resources

packages/core/src/agent/DextoAgent.ts
```diff
@@ -345,7 +349,17 @@ export class DextoAgent {
...
+            let finalText = textInput;
+            if (textInput && textInput.includes('@')) {
+                const resources = await this.resourceManager.list();
+                const expansion = await expandMessageReferences(textInput, resources, (uri) =>
+                    this.resourceManager.read(uri)
+                );
+                finalText = expansion.expandedMessage;
+            }
+
+            const response = await session.run(finalText, imageDataInput, fileDataInput, stream);
```
rahulkarajgikar, line 362: be careful with this

not all @ mentions may be resources, it could just be a way user is trying to talk to the agent

cursor has this really annoying feature where if you put an @, it doesn't let you escape their resources menu, even if you aren't trying to use a resource (sometimes you might jsut want to use the @ character normally). even typing space doesn't exit their 'resources' menu

i've seen similar behaviour in some other tools and the best UX i've seen is if you type space, it escapes the resources/popup menu and let's you continue your prompt assuming @ is a normal character- github has this feature (try typing @ in github)

haven't checked the exact impl to see if this works that way but just a thought

packages/webui/components/types/resources.ts
```diff
@@ -0,0 +1,11 @@
...
+    uri: string;
+    name?: string;
+    description?: string;
+    mimeType?: string;
+    source: 'mcp' | 'plugin' | 'custom';
+    serverName?: string;
+    size?: number;
+    lastModified?: string | Date;
+    metadata?: Record<string, unknown>;
+};
```
rahulkarajgikar, line 11: import the type from core, libraries are properly separated now

limit duplication as much as possible

rahulkarajgikar (changes requested):
reviewed

need to understand more what the future scope of resources would look like to know if our current abstractions are right or necessary - added comments in review

(haven't reviewed UI code or tested impl)

other things to do:
- add changeset
- add docstrings to important classes and functions 
- add unit tests for parsing functions and util functions at least (defer other tests for later once implementation is decided)
- update documentation (can do this in follow up)

Issue #315, opened 2025-09-02
Clean up sanitization code and helpers


shaunak99:
Sanitization logic and utilities need to be standardizes and clean up for media tool results. This should be addressed and combined with a file storage layer to create a unified and standardized sanitization, parsing and referencing layer.

Issue #87, opened 2025-05-01
Enhance MCP client functionality to support roots & sampling


shaunak99:
Read https://github.com/modelcontextprotocol/typescript-sdk/blob/854b55e0a1783e757c2e3dcfb9d088db31cc9821/src/client/index.ts

use the code in this + some experimentation with the MCP APIs to implement functions for roots and sampling in our mcp client.

PR #355, opened 2025-09-14
Adding blob store, ResourceManager, PromptsManager

shaunak99:
This PR introduces a blob store and builds on top of the ResourceManager from #354 to consolidate all the blob data handling. Additionally, we have migrated #350 to this PR along with addressing #87.

Regarding toolResult sanitization in #315 :

  - OpenAI/Anthropic: every tool run immediately calls ContextManager.addToolResult(...). That helper performs the blob-aware sanitization once, logs it, and appends a tool message to the running history before the next model turn starts.
  - Vercel: the SDK asks us for an inline return value from the tool execute handler each time it pauses. That value becomes the toolResult the SDK emits back to the model, and the full conversation transcript—including those tool results—arrives only after the SDK finishes the multi-step interaction. When that final transcript comes back, VercelMessageFormatter.parseResponse(...) converts it into internal messages and ContextManager.processLLMResponse(...) adds them to history in one batch. If we tried to push tool results into the history from inside execute, we’d end up double-logging the same result (once immediately, once again when the final transcript is applied). Today we already share the routines (sanitizeToolResultToContent(WithBlobs) plus summarizeToolContentForText) between the two flows; the difference is just when we run them.

  Because of that SDK contract, there isn’t a clean way to unify the control flow without major surgery—e.g. intercepting and stripping tool results from the post-run transcript, or reworking the Vercel transport to drive history updates itself. Both approaches add more complexity than they remove. So the current split (immediate addToolResult for “live update” providers vs. sanitizing/queuing inside the Vercel tool handler and letting the final transcript update the session) is intentional, and keeps each integration correct with minimal duplication.

## Summary by CodeRabbit

* **New Features**
  * Slash-command prompt UI, Create Prompt modal, starter prompts as quick actions.
  * Resource & media support: @-mention autocomplete, searchable resource picker, inline previews (images/audio/files), resource-aware message rendering.
  * Blob-backed file storage: upload, store, list, retrieve, size limits, cleanup; prompts may include attached resources.
  * Live updates: real-time prompt and resource refreshes across UI via websocket events.

* **Documentation**
  * Added MCP Resources demo and new prompt guides.

* **Chores**
  * Dependency bumps and build-script updates.


packages/core/src/agent/DextoAgent.ts
```diff
@@ -517,7 +588,17 @@ export class DextoAgent {
...
+        if (!this.resourceManager) {
+            return history;
+        }
+
+        return await Promise.all(
+            history.map(async (message) => ({
+                ...message,
+                content: await expandBlobReferences(message.content, this.resourceManager),
+            }))
+        );
```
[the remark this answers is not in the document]
shaunak99: Good catch!

packages/core/src/prompts/providers/custom-prompt-provider.ts
```diff
@@ -0,0 +1,319 @@
...
+                logger.warn(
+                    `Failed to load blob resource for custom prompt ${name}: ${String(error)}`
+                );
+            }
+        }
+
+        return {
+            description: record.description,
+            messages,
+        };
```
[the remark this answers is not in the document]
shaunak99: internal-prompt-provider.ts also uses single content objects (see lines showing content: resourceContent). This is an established pattern in the codebase, can you double check?

agents/default-agent.yml
```diff
@@ -41,24 +34,83 @@ systemPrompt:
...
+    title: "Connect New Tools"
+    description: "Browse and add MCP servers to extend capabilities"
+    prompt: "I want to connect new tools to expand my capabilities. Can you help me understand what MCP servers are available and how to add them?"
+    category: tools
+    icon: "🔧"
+    priority: 3
+
+# Blob storage configuration - infrastructure-level storage for large files
+# This section controls the actual blob backend, size limits, and cleanup policies
+blobStorage:
```
rahulkarajgikar, line 104: should move to existing storage schema, along with code moving to storage module:

```
storage
 - cache: <<>
 - database: <>
 - blob: <>
```

agents/default-agent.yml
```diff
@@ -41,24 +34,83 @@ systemPrompt:
...
+      maxFiles: 50               # Maximum number of files to index
+      maxDepth: 3                # Maximum directory depth to traverse
+      includeHidden: false         # Include hidden files/directories
+      includeExtensions: [".txt", ".md", ".json", ".yaml", ".yml", ".js", ".ts", ".py", ".html", ".css"]
+
+    # Blob resource - enables large file upload/storage (settings in blobStorage section)
+    - type: blob
+
+# Starter prompts - predefined prompts that appear as clickable buttons in the WebUI
+starterPrompts:
```
rahulkarajgikar, line 72: have to update docs for this starterPrompts and internalResources section and blob section

examples/resources-demo-server/agent.yml
```diff
@@ -0,0 +1,30 @@
+# MCP Resources Demo Agent Configuration
+# This agent demonstrates MCP Resources capability from external server
+
+# MCP Servers configuration
+mcpServers:
+  resources-demo:
+    type: stdio
+    command: node
+    args:
+      - "./examples/resources-demo-server/server.js"
```
rahulkarajgikar, line 10: use @dexto.agent_dir path, not relative path to avoid this breaking based on run location. avoid any relative paths in yml files 


right path: `@dexto.agent_dir/server.js` (check exact syntax from registry agents)

examples/resources-demo-server/agent.yml
```diff
@@ -0,0 +1,30 @@
...
+    timeout: 30000
+    connectionMode: lenient
+
+# LLM configuration
+llm:
+  provider: openai
+  model: gpt-4o-mini
```
rahulkarajgikar, line 17: use gpt-5-mini

packages/cli/src/api/mcp/tool-aggregation-handler.ts
```diff
@@ -32,52 +33,39 @@ export async function initializeMcpToolAggregationServer(
...
-        const mcpTools = await connectedClient.listTools({});
 
-        logger.debug(`MCP client name: ${clientName}`);
-        logger.debug(`MCP client tools: ${JSON.stringify(mcpTools, null, 2)}`);
+    for (const [toolName, toolDef] of Object.entries(toolDefinitions)) {
+        toolCount++;
+        const jsonSchema = toolDef.parameters ?? { type: 'object', properties: {} };
+        const paramsShape = jsonSchemaToZodShape(jsonSchema);
+        const _paramsSchema = z.object(paramsShape);
+        type ToolArgs = z.infer<typeof _paramsSchema>;
```
rahulkarajgikar, line 44: nit: use z.output. add linter rule for enforcing this

packages/cli/src/api/mcp/tool-aggregation-handler.ts
```diff
@@ -32,52 +33,39 @@ export async function initializeMcpToolAggregationServer(
...
-        for (const tool of mcpTools.tools) {
-            toolCount++;
-            logger.debug(`Registering tool: ${tool.name}`);
-
-            // Convert JSON Schema to Zod raw shape
-            const zodShape = jsonSchemaToZodShape(tool.inputSchema);
-
-            // Log the tool schema to debug the issue (safely handle Zod schema circular references)
+        const level = typeof logger.getLevel === 'function' ? logger.getLevel() : 'info';
+        if (level === 'debug' || level === 'verbose' || level === 'silly') {
```
rahulkarajgikar, line 47: if condition is not necessary, logger.debug already handles this and only shows if log level matches the above

packages/cli/src/api/mcp/tool-aggregation-handler.ts
```diff
@@ -87,19 +75,13 @@ export async function initializeMcpToolAggregationServer(
...
-            );
-            resourceIndex++;
-        }
+        allResources.forEach((resource, index) => {
+            const safeId = resource.key.replace(/[^a-zA-Z0-9]/g, '_');
+            mcpServer.resource(`resource_${index}_${safeId}`, resource.key, async () => {
+                logger.info(`Resource aggregation: reading ${resource.key}`);
+                return await mcpManager.readResource(resource.key);
+            });
+        });
```
rahulkarajgikar, line 84: note: verify if client collisions are handled and re-test this aggregator flow

packages/cli/src/api/server.ts
```diff
@@ -272,11 +300,160 @@ export async function initializeApi(
...
+        return new DextoValidationError([
+            {
+                code: AgentErrorCode.API_VALIDATION_ERROR,
+                message: errorMessage,
+                scope: ErrorScope.AGENT,
+                type: ErrorType.USER,
+                severity: 'error' as const,
+            },
+        ]);
+    }
```
rahulkarajgikar, line 316: this func is not necessary, middleware handles this (verify this)

packages/cli/src/api/server.ts
```diff
@@ -272,11 +300,160 @@ export async function initializeApi(
...
+
+    function decodeResourceId(resourceIdParam: string): string {
```
rahulkarajgikar, line 318: refactor: this is a function that has one line and calls another function. not needed, move this logic into decodeURIComponent if necessary and remove this function

packages/cli/src/api/server.ts
```diff
@@ -272,11 +300,160 @@ export async function initializeApi(
...
+        }
+    });
+
+    // Get a specific prompt definition
+    app.get('/api/prompts/:name', async (req, res, next) => {
+        try {
+            const name = req.params.name;
+            if (!name) throw PromptError.nameRequired();
+            const definition = await activeAgent.getPromptDefinition(name);
+            if (!definition) throw PromptError.notFound(name);
```
rahulkarajgikar, line 402: avoid manually checking each argument, use zod schemas and parseBody everywhere for consistency

packages/cli/src/api/server.ts
```diff
@@ -272,11 +300,160 @@ export async function initializeApi(
...
+            // Build options object with only defined values (exactOptionalPropertyTypes compatibility)
+            const options: {
+                q?: string;
+                context?: string;
+                args?: Record<string, unknown>;
+            } = {};
+            if (q !== undefined) options.q = q;
+            if (ctx !== undefined) options.context = ctx;
+            if (parsedArgs !== undefined) options.args = parsedArgs;
+
```
rahulkarajgikar, line 444: avoid manually checking each argument, use zod schemas and parseBody everywhere for consistency

packages/cli/src/api/server.ts
```diff
@@ -533,6 +710,132 @@ export async function initializeApi(
...
+        } catch (error) {
+            if (error instanceof z.ZodError)
+                return next(handleZodError(error, 'Invalid resourceId validation'));
+            const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+            logger.error(
+                `Error reading resource content for '${decodedResourceId}': ${errorMessage}`
+            );
+            return next(error);
+        }
+    });
```
rahulkarajgikar, line 750: avoid manually checking each argument, use zod schemas and parseBody everywhere for consistency

packages/cli/src/api/server.ts
```diff
@@ -533,6 +710,132 @@ export async function initializeApi(
...
+                        type: ErrorType.USER,
+                        severity: 'error' as const,
+                        context: { resourceIdParam },
+                    },
+                ])
+            );
+        }
+        try {
+            const qualifiedUri = `mcp:${serverId}:${decodedResourceId}`;
+            const content = await activeAgent.readResource(qualifiedUri);
```
rahulkarajgikar, line 828: avoid manually checking each argument, use zod schemas and parseBody everywhere for consistency

packages/cli/src/api/server.ts
```diff
@@ -533,6 +710,132 @@ export async function initializeApi(
...
+            const exists = await activeAgent.hasResource(validatedResourceId);
+            return res.status(exists ? 200 : 404).end();
+        } catch (error) {
+            if (error instanceof z.ZodError)
+                return next(handleZodError(error, 'Invalid resourceId validation'));
+            const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+            logger.error(
+                `Error checking resource existence for '${decodedResourceId}': ${errorMessage}`
+            );
+            return next(error);
```
rahulkarajgikar, line 774: avoid manually checking each argument, use zod schemas and parseBody everywhere for consistency

could potentially use zod schema transforms for this resource id decoding which is happening very frequently

packages/cli/src/cli/commands/interactive-commands/prompt-commands.ts
```diff
@@ -38,4 +42,308 @@ export const promptCommands: CommandDefinition[] = [
...
         },
     },
+    {
+        name: 'prompts',
+        description: 'List all available prompts (MCP + internal)',
+        usage: '/prompts',
+        category: 'Prompt Management',
+        handler: async (_args: string[], agent: DextoAgent): Promise<boolean> => {
+            try {
+                const prompts = (await agent.promptsManager.list()) as Record<string, PromptInfo>;
```
rahulkarajgikar, line 52: verify if casting is necessary

packages/cli/src/cli/commands/interactive-commands/prompt-commands.ts
```diff
@@ -38,4 +42,308 @@ export const promptCommands: CommandDefinition[] = [
...
+                    chalk.red(
+                        `Error listing prompts: ${error instanceof Error ? error.message : String(error)}`
+                    )
+                );
+                return false;
+            }
+        },
+    },
+    {
+        name: 'use',
```
rahulkarajgikar, line 159: nit: rename to 'use prompt'?

packages/cli/src/cli/commands/interactive-commands/prompt-commands.ts
```diff
@@ -38,4 +42,308 @@ export const promptCommands: CommandDefinition[] = [
...
+        },
+    };
+}
+
+/**
+ * Get all dynamic prompt commands based on available prompts
+ */
+export async function getDynamicPromptCommands(agent: DextoAgent): Promise<CommandDefinition[]> {
+    try {
+        const prompts = (await agent.promptsManager.list()) as Record<string, PromptInfo>;
```
rahulkarajgikar, line 307: nit: check if casting is necssary

packages/cli/package.json
```diff
@@ -28,7 +28,7 @@
     "zod": "^3.25.0"
   },
   "scripts": {
-    "build": "tsc -p tsconfig.build.json && pnpm run copy-agents",
+    "build": "cross-env NODE_OPTIONS='--max-old-space-size=4096' tsc -p tsconfig.build.json && pnpm run copy-agents",
```
rahulkarajgikar, line 31: TODO: this is a known issue with MCP sdk >= 1.18.1 taking heap space, see if we can solve this without heap size hacks

packages/core/src/agent/DextoAgent.ts
```diff
@@ -1003,6 +1105,59 @@ export class DextoAgent {
         return this.mcpManager.getFailedConnections();
     }
 
+    // ============= RESOURCE MANAGEMENT =============
+
+    /**
+     * Lists all available resources with their info.
+     * This includes resources from MCP servers and any custom resource providers.
+     */
+    public async listResources(): Promise<import('../resources/index.js').ResourceSet> {
```
rahulkarajgikar, line 1114: nit: verify that usages of resource/prompt methods in CLI are using dextoagent resource/prompt methods rather than directly accessing promptmanager/resource manager through the agent

packages/core/src/agent/DextoAgent.ts
```diff
@@ -1252,3 +1502,5 @@ export class DextoAgent {
     // - Tool chaining and workflow automation
     // - Agent collaboration and delegation
 }
+
+// (resource methods are defined on the class above; no interface augmentation required)
```
rahulkarajgikar, line 1506: nit: remove stray comment

packages/core/src/agent/schemas.ts
```diff
@@ -114,6 +116,82 @@ export const AgentConfigSchema = z
...
+            .default([])
+            .describe('Starter prompts that appear as clickable buttons in the WebUI'),
+
+        // Blob storage configuration (infrastructure-level blob storage)
+        blobStorage: BlobServiceConfigSchema.default({
+            type: 'local',
+            maxBlobSize: 50 * 1024 * 1024, // 50MB
+            maxTotalSize: 1024 * 1024 * 1024, // 1GB
+            cleanupAfterDays: 30,
+        }).describe('Blob storage backend configuration for large file handling'),
```
rahulkarajgikar, line 194: move into storage schema as a sub-schema along with all blob store code being moved into storage folder

also add .strict()

packages/core/src/agent/schemas.ts
```diff
@@ -114,6 +116,82 @@ export const AgentConfigSchema = z
...
+                        category: z
+                            .enum(['general', 'coding', 'analysis', 'tools', 'learning'])
+                            .optional()
+                            .default('general')
+                            .describe('Category for organizing starter prompts'),
+                        icon: z
+                            .string()
+                            .optional()
+                            .default('')
+                            .describe('Emoji or icon to display'),
```
rahulkarajgikar, line 159: nit: remove this and keep it in descrpition

packages/core/src/agent/schemas.ts
```diff
@@ -114,6 +116,82 @@ export const AgentConfigSchema = z
...
+                            .default('')
+                            .describe('Description shown on hover or in the UI'),
+                        prompt: z
+                            .string()
+                            .describe('The actual prompt text that gets resolved and sent'),
+                        category: z
+                            .enum(['general', 'coding', 'analysis', 'tools', 'learning'])
+                            .optional()
+                            .default('general')
+                            .describe('Category for organizing starter prompts'),
```
rahulkarajgikar, line 154: nit: might be better to not use a strict enum here and be more open-ended to allow any type of tags

packages/core/src/blob/backend/local-backend.ts
```diff
@@ -0,0 +1,588 @@
...
+ * Internal configuration with all required properties
+ */
+interface ResolvedLocalBlobConfig {
+    type: 'local';
+    maxBlobSize: number;
+    maxTotalSize: number;
+    cleanupAfterDays: number;
+    storePath: string;
+}
+
```
rahulkarajgikar, line 29: dup: derive from zod schema

packages/core/src/blob/schemas.ts
```diff
@@ -0,0 +1,103 @@
...
+    })
+    .strict()
+    .describe('Storage statistics');
+
+// Type exports for convenience
+export type ValidatedBlobServiceConfig = z.infer<typeof BlobServiceConfigSchema>;
+export type ValidatedBlobMetadata = z.infer<typeof BlobMetadataSchema>;
+export type ValidatedStoredBlobMetadata = z.infer<typeof StoredBlobMetadataSchema>;
+export type ValidatedBlobReference = z.infer<typeof BlobReferenceSchema>;
+export type ValidatedBlobStats = z.infer<typeof BlobStatsSchema>;
```
rahulkarajgikar, line 103: nit: use z.output, delete unused schemas

packages/core/src/blob/blob-service.ts
```diff
@@ -0,0 +1,201 @@
...
+    BlobData,
+    BlobStats,
+} from './types.js';
+
+/**
+ * Main blob service implementation for local filesystem storage
+ *
+ * Provides blob storage operations using the local filesystem backend.
+ */
+export class BlobService implements IBlobService {
```
rahulkarajgikar, line 20: todo: see how this can be integrated into existing storage modules and what needs to be refactored for that

notes:
- we likely don't need this class and can use storage-manager
- error codes don't need blob scope, but storage scope
- new backends would move into storage/backend
- new schemas move into storage/schemas and types in storage/types

packages/core/src/blob/backend/local-backend.ts
```diff
@@ -0,0 +1,588 @@
...
+ */
+export class LocalBlobBackend implements BlobBackend {
+    private config: ResolvedLocalBlobConfig;
+    private storePath: string;
+    private connected = false;
+    private statsCache: { count: number; totalSize: number } | null = null;
+    private statsCachePromise: Promise<void> | null = null;
+    private lastStatsRefresh: number = 0;
+
+    private static readonly DEFAULT_CONFIG = {
```
rahulkarajgikar, line 45: avoid defaults here and fallback defaults in constructor, zod would have them so the config already has everything it needs by this point

packages/core/src/blob/backend/local-backend.ts
```diff
@@ -0,0 +1,588 @@
...
+}
+
+/**
+ * Local filesystem blob backend implementation
+ *
+ * Stores blobs on the local filesystem with content-based deduplication
+ * and metadata tracking. This is the default backend for development
+ * and single-machine deployments.
+ */
+export class LocalBlobBackend implements BlobBackend {
```
rahulkarajgikar, line 37: for agent: review this implementation for local filesystem

packages/core/src/blob/schemas.ts
```diff
@@ -0,0 +1,103 @@
+import { z } from 'zod';
+
+/**
```
rahulkarajgikar, line 3: agent: check this file for unused schemas and delete them

packages/core/src/blob/types.ts
```diff
@@ -0,0 +1,198 @@
+/**
+ * Core types for the Blob Service system
```
rahulkarajgikar, line 2: agent: check this file for dup types that can be derived from schema

prompts/code-review.md
```diff
@@ -0,0 +1,101 @@
+---
+description: "Review code for bugs, improvements, and best practices with actionable feedback"
+id: code-review
+name: Code Review Assistant
+command: /code-review
+category: coding
+---
+
```
rahulkarajgikar, line 8: TODO: move all prompts off absolute path and into relative agent specific paths, referenced using @dexto.agent_dir colocated near their folders

this allows us to keep agent specific prompts

packages/core/src/context/manager.ts
```diff
@@ -84,38 +90,139 @@ export class ContextManager<TMessage = unknown> {
     private historyProvider: IConversationHistoryProvider;
     private readonly sessionId: string;
 
+    /**
+     * Optional ResourceManager for resolving blob references in message content.
+     * When provided, blob references like @blob:abc123 will be resolved to actual data
+     * before passing messages to the LLM formatter.
+     */
+    private resourceManager?: import('../resources/index.js').ResourceManager | undefined;
```
rahulkarajgikar, line 98: agent: check if this should be optional or mandatory (i guess mandatory)

packages/core/src/context/manager.ts
```diff
@@ -84,38 +90,139 @@ export class ContextManager<TMessage = unknown> {
...
         maxInputTokens: number,
         tokenizer: ITokenizer,
         historyProvider: IConversationHistoryProvider,
         sessionId: string,
         compressionStrategies: ICompressionStrategy[] = [
             new MiddleRemovalStrategy(),
             new OldestRemovalStrategy(),
-        ]
+        ],
+        resourceManager?: import('../resources/index.js').ResourceManager
```
rahulkarajgikar, line 215: agent: check if this should be mandatory

packages/core/src/errors/types.ts
```diff
@@ -26,6 +28,8 @@ export enum ErrorScope {
     SYSTEM_PROMPT = 'system_prompt', // System prompt contributors and file processing
     PREFERENCE = 'preference', // Global preferences file operations and validation
     AGENT_REGISTRY = 'agent_registry', // Agent registry operations, installation, resolution
+    RESOURCE = 'resource', // Resource management (MCP/internal) discovery and access
+    BLOB = 'blob', // Blob storage operations and backend management
```
rahulkarajgikar, line 32: delete this once blob is moved into storage layer, same as below

packages/core/src/llm/services/factory.ts
```diff
@@ -138,21 +143,23 @@ function getOpenAICompatibleBaseURL(llmConfig: ValidatedLLMConfig): string {
 function _createVercelLLMService(
     config: ValidatedLLMConfig,
     toolManager: ToolManager,
-    promptManager: PromptManager,
+    systemPromptManager: SystemPromptManager,
     historyProvider: IConversationHistoryProvider,
     sessionEventBus: SessionEventBus,
-    sessionId: string
+    sessionId: string,
+    resourceManager?: import('../../resources/index.js').ResourceManager
```
rahulkarajgikar, line 150: make resourceManager argument mandatory everywhere. it's always defined

packages/core/src/llm/services/factory.ts
```diff
@@ -163,28 +170,31 @@ export function createLLMService(
     config: ValidatedLLMConfig,
     router: LLMRouter,
     toolManager: ToolManager,
-    promptManager: PromptManager,
+    systemPromptManager: SystemPromptManager,
     historyProvider: IConversationHistoryProvider,
     sessionEventBus: SessionEventBus,
-    sessionId: string
+    sessionId: string,
+    resourceManager?: import('../../resources/index.js').ResourceManager
```
rahulkarajgikar, line 177: mandatory

packages/core/src/llm/services/openai.ts
```diff
@@ -31,11 +31,12 @@ export class OpenAIService implements ILLMService {
...
         toolManager: ToolManager,
         openai: OpenAI,
-        promptManager: PromptManager,
+        systemPromptManager: SystemPromptManager,
         historyProvider: IConversationHistoryProvider,
         sessionEventBus: SessionEventBus,
         config: ValidatedLLMConfig,
-        sessionId: string
+        sessionId: string,
+        resourceManager?: import('../../resources/index.js').ResourceManager
```
rahulkarajgikar, line 39: mandatory

packages/core/src/llm/services/vercel.ts
```diff
@@ -52,11 +56,12 @@ export class VercelLLMService implements ILLMService {
...
         toolManager: ToolManager,
         model: LanguageModel,
-        promptManager: PromptManager,
+        systemPromptManager: SystemPromptManager,
         historyProvider: IConversationHistoryProvider,
         sessionEventBus: SessionEventBus,
         config: ValidatedLLMConfig,
-        sessionId: string
+        sessionId: string,
+        resourceManager?: import('../../resources/index.js').ResourceManager
```
rahulkarajgikar, line 64: mandatory

packages/core/src/context/utils.ts
```diff
@@ -287,11 +639,312 @@ function sanitizeDeepObject(obj: unknown): unknown {
...
+    }
+}
+
 /**
  * Convert an arbitrary tool result into safe InternalMessage content.
  * - Converts data URIs and base64 blobs to media/file parts
  * - Removes huge binary blobs inside objects
  * - Truncates extremely long raw text
+ *
+ * @deprecated Use sanitizeToolResultToContentWithBlobs for automatic blob storage
```
rahulkarajgikar, line 947: check for usages and delete, prefer to avoid @deprecated in early stages of project

packages/core/src/context/utils.ts
```diff
@@ -11,6 +11,152 @@ const DEFAULT_OVERHEAD_PER_MESSAGE = 4; // Approximation for message format over
 const MIN_BASE64_HEURISTIC_LENGTH = 512; // Below this length, treat as regular text
 const MAX_TOOL_TEXT_CHARS = 8000; // Truncate overly long tool text
 
+type ToolBlobNamingOptions = {
```
rahulkarajgikar, line 14: agent: add tests for utility functions in this file, check for duplicates, unused functions and delete them

packages/core/src/context/manager.ts
```diff
@@ -444,8 +570,14 @@ export class ContextManager<TMessage = unknown> {
         }
         // Sanitize tool result to avoid adding non-text data as raw text
         // and to convert media/data-uris/base64 to structured parts.
-        let content: InternalMessage['content'];
-        content = sanitizeToolResultToContent(result);
+        // If a resource manager is available, automatically store large media as blobs.
+        const blobService = this.resourceManager?.getBlobService();
+        const content = blobService
```
rahulkarajgikar, line 575: make blobService also mandatory and avoid optional checks, fallbacks

packages/core/src/context/manager.ts
```diff
@@ -501,9 +633,23 @@ export class ContextManager<TMessage = unknown> {
     ): Promise<TMessage[]> {
         // TMessage type is provided by the service that instantiates ContextManager
         // Use provided history or fetch from provider
-        const messageHistory: InternalMessage[] =
+        let messageHistory: InternalMessage[] =
             history ?? (await this.historyProvider.getHistory());
 
+        // Resolve blob references if resource manager is available
+        if (this.resourceManager) {
```
rahulkarajgikar, line 640: drop if after making resource manager mandatory

packages/core/src/mcp/manager.ts
```diff
@@ -397,33 +454,47 @@ export class MCPManager {
...
 
     /**
-     * Get all available resource URIs from all connected clients, updating the cache.
-     * @returns Promise resolving to an array of unique resource URIs.
+     * Get all cached MCP resources (no network calls).
      */
-    async listAllResources(): Promise<string[]> {
-        return Array.from(this.resourceToClientMap.keys());
+    async listAllResources(): Promise<MCPResolvedResource[]> {
+        return Array.from(this.resourceCache.values()).map(({ key, serverName, summary }) => ({
```
rahulkarajgikar, line 460: might need to check if cache is up to date and update if necessary

similar for other methods

packages/core/src/mcp/mcp-client.ts
```diff
@@ -487,4 +491,48 @@ export class MCPClient implements IMCPClient {
...
+
+    /**
+     * Set up notification handlers for MCP server notifications
+     */
+    private setupNotificationHandlers(): void {
+        if (!this.client) return;
+
+        try {
+            // Resource updated
+            (this.client as any).setNotificationHandler?.(
```
rahulkarajgikar, line 503: avoid any cast

packages/core/src/mcp/mcp-client.ts
```diff
@@ -487,4 +491,48 @@ export class MCPClient implements IMCPClient {
...
+                'notifications/resources/updated',
+                (params: { uri: string; title?: string }) => this.handleResourceUpdated(params)
+            );
+        } catch (error) {
+            logger.warn(`Could not set resources/updated notification handler: ${error}`);
+        }
+        try {
+            // Prompts list changed
+            (this.client as any).setNotificationHandler?.(
```
rahulkarajgikar, line 512: avoid any cast

packages/core/src/mcp/mcp-client.ts
```diff
@@ -397,12 +396,17 @@ export class MCPClient implements IMCPClient {
...
      * TODO: Turn exception logs back into error and only call this based on capabilities of the server
      */
-    async listResources(): Promise<string[]> {
+    async listResources(): Promise<MCPResourceSummary[]> {
         this.ensureConnected();
         try {
             const response = await this.client!.listResources();
             logger.debug(`listResources response: ${JSON.stringify(response, null, 2)}`);
-            return response.resources.map((r: any) => r.uri);
+            return response.resources.map((r: any) => ({
```
rahulkarajgikar, line 404: avoid any

packages/core/src/mcp/manager.ts
```diff
@@ -37,14 +38,21 @@ import { MCPError } from './errors.js';
  * const tools = await manager.getAllTools();
  * ```
  */
+type ResourceCacheEntry = {
+    key: string;
```
rahulkarajgikar, line 42: this key is likely duplicate of the key in the map

packages/core/src/mcp/manager.ts
```diff
@@ -581,8 +652,117 @@ export class MCPManager {
...
+        }
+    }
+
+    /**
+     * Handle prompts list changed notification
+     */
+    private async handlePromptsListChanged(serverName: string, client: IMCPClient): Promise<void> {
+        try {
+            // Refresh the prompts for this client
+            const existingPrompts = Array.from(this.promptToClientMap.entries())
```
rahulkarajgikar, line 738: similar promptCache like the new resourceCache might make sense

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,379 @@
...
+
+interface PromptCacheEntry {
+    key: string;
+    providerName: string;
+    providerPromptName: string;
+    originalName: string;
+    info: PromptInfo;
+}
+
+export class PromptsManager {
```
rahulkarajgikar, line 26: nit: rename to PromptManager for consistency with other services + add docstring

packages/core/src/resources/core.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+export type { ResourceSource, ResourceMetadata, ResourceProvider, ResourceSet } from './types.js';
+
+export type { InternalResourcesConfig, ValidatedInternalResourcesConfig } from './schemas.js';
+
+export { ResourceManager } from './manager.js';
+
+export { ResourceError } from './errors.js';
+export { ResourceErrorCodes } from './error-codes.js';
```
rahulkarajgikar, line 12: avoid 'core' files. don't introduce new inconsistent paradigms for exports. stick to index, types.ts which we already have everywhere

packages/core/src/prompts/providers/starter-prompt-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+    invalidateCache(): void {
+        this.cacheValid = false;
+        this.promptsCache = [];
+        logger.debug('StarterPromptProvider cache invalidated');
+    }
+
+    /**
+     * Update starter prompts configuration
+     */
+    updateConfig(agentConfig?: ValidatedAgentConfig): void {
```
rahulkarajgikar, line 47: unnecessary optional argument

packages/core/src/prompts/providers/starter-prompt-provider.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Starter prompts are intended for quick access to common user workflows and are typically
+ * displayed prominently in the UI.
+ */
+export class StarterPromptProvider implements PromptProvider {
+    // Sourced from validated AgentConfig schema's starterPrompts
+    private starterPrompts: StarterPromptItem[] = [];
+    private promptsCache: PromptInfo[] = [];
+    private cacheValid: boolean = false;
+
+    constructor(agentConfig?: ValidatedAgentConfig) {
```
rahulkarajgikar, line 22: unnecessary optional argument

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,379 @@
...
+    private aliasMap: Map<string, string> = new Map();
+    private buildPromise: Promise<void> | null = null;
+
+    constructor(
+        mcpManager: MCPManager,
+        resourceManager: ResourceManager,
+        promptsDir?: string,
+        agentConfig?: ValidatedAgentConfig,
+        private readonly eventBus?: AgentEventBus,
+        private readonly database?: DatabaseBackend
```
rahulkarajgikar, line 38: unneceessary optional args

packages/core/src/prompts/providers/custom-prompt-provider.ts
```diff
@@ -0,0 +1,319 @@
...
+}
+
+export class CustomPromptProvider implements PromptProvider {
+    private cacheValid = false;
+    private promptsCache: PromptInfo[] = [];
+    private promptRecords: Map<string, StoredCustomPrompt> = new Map();
+
+    constructor(
+        private database: DatabaseBackend,
+        private resourceManager?: ResourceManager
```
rahulkarajgikar, line 52: unnecessary optional

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,379 @@
...
+    CustomPromptProvider,
+    type CreateCustomPromptInput,
+} from './providers/custom-prompt-provider.js';
+import { PromptError } from './errors.js';
+import { logger } from '../logger/index.js';
+import type { ResourceManager } from '../resources/manager.js';
+import type { DatabaseBackend } from '../storage/backend/database-backend.js';
+
+interface PromptCacheEntry {
+    key: string;
```
rahulkarajgikar, line 19: likely duplicate key. we also might be doing double caching in both prompt manager and the providers which is unnecessary

packages/core/src/prompts/providers/internal-prompt-provider.ts
```diff
@@ -0,0 +1,319 @@
...
+    promptsDir?: string;
+    resourceManager: ResourceManager;
+}
+
+interface ParsedPrompt {
+    info: PromptInfo;
+    content: string;
+}
+
+export class InternalPromptProvider implements PromptProvider {
```
rahulkarajgikar, line 19: might not actually need INternalPromptProvider, seems equivalent to starter-prompt-provider with a hardcoded directory. can keep for now but basically we can add file based prompt providers instead (refer to relative file colocated near the agent)

packages/core/src/prompts/utils.ts
```diff
@@ -0,0 +1,141 @@
+import type { GetPromptResult } from '@modelcontextprotocol/sdk/types.js';
```
rahulkarajgikar, line 1: agent: add tests for util functions in this file

packages/core/src/prompts/prompts-manager.ts
```diff
@@ -0,0 +1,379 @@
...
+        }
+        this.buildPromise = this.buildCache();
+        try {
+            await this.buildPromise;
+        } finally {
+            this.buildPromise = null;
+        }
+    }
+
+    private async buildCache(): Promise<void> {
```
rahulkarajgikar, line 271: agent: check if double caching in prompt manager and providrs can be simplified, add a TODO for this at least

packages/core/src/resources/reference.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Resource reference parsing and expansion
+ */
+
+export {
+    parseResourceReferences,
+    resolveResourceReferences,
+    expandMessageReferences,
+    formatResourceContent,
+} from './reference-parser.js';
```
rahulkarajgikar, line 10: delete this file

packages/core/src/resources/internal-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+                } catch (error) {
+                    logger.error(
+                        `Failed to read resource ${uri} from ${type} handler: ${error instanceof Error ? error.message : String(error)}`,
+                        error
+                    );
+                    throw error;
+                }
+            }
+        }
+        throw new Error(`No handler found for resource: ${uri}`);
```
rahulkarajgikar, line 88: throw a typed error instead

packages/core/src/resources/internal-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export class InternalResourcesProvider implements ResourceProvider {
+    private config: ValidatedInternalResourcesConfig;
+    private handlers: Map<string, InternalResourceHandler> = new Map();
+    private services: InternalResourceServices;
+
+    constructor(config: ValidatedInternalResourcesConfig, services: InternalResourceServices = {}) {
+        this.config = config;
+        this.services = services;
+        logger.debug('InternalResourcesProvider initialized with config:', config);
```
rahulkarajgikar, line 21: nit: incorrect use of logger args

packages/core/src/resources/internal.ts
```diff
@@ -0,0 +1,19 @@
...
+} from './internal-registry.js';
+export type {
+    InternalResourceConfig,
+    InternalResourceHandler,
+    InternalResourceServices,
+    BlobResourceConfig,
+} from './internal-registry.js';
+
+// Legacy BlobStore has been fully migrated to BlobService
+// All blob storage functionality is now provided by the infrastructure-level BlobService
```
rahulkarajgikar, line 19: delete this file

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,764 @@
+import { ResourceMetadata } from './types.js';
```
rahulkarajgikar, line 1: nit: refactor this file into separate implementations - filesystem resource handler and blobresource handler

packages/core/src/resources/manager.ts
```diff
@@ -0,0 +1,235 @@
...
+import { InternalResourcesProvider } from './internal-provider.js';
+import type { ReadResourceResult } from '@modelcontextprotocol/sdk/types.js';
+import type { ValidatedInternalResourcesConfig } from './schemas.js';
+import type { InternalResourceServices } from './internal-registry.js';
+import { logger } from '../logger/index.js';
+import { ResourceError } from './errors.js';
+import { eventBus } from '../events/index.js';
+
+export interface ResourceManagerOptions {
+    internalResourcesConfig?: ValidatedInternalResourcesConfig;
```
rahulkarajgikar, line 12: unnecessary optional config

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,764 @@
...
+            '.tex': 'text/x-tex',
+            '.dockerfile': 'text/x-dockerfile',
+        };
+        return mimeTypes[ext] || 'text/plain';
+    }
+}
+
+export class BlobResourceHandler implements InternalResourceHandler {
+    private config?: BlobResourceConfig;
+    private blobService?: BlobService;
```
rahulkarajgikar, line 490: incorrect optionals

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,764 @@
...
+    getType(): string;
+    initialize(config: InternalResourceConfig, services: InternalResourceServices): Promise<void>;
+    listResources(): Promise<ResourceMetadata[]>;
+    readResource(uri: string): Promise<ReadResourceResult>;
+    canHandle(uri: string): boolean;
+    refresh?(): Promise<void>;
+}
+
+export class FileSystemResourceHandler implements InternalResourceHandler {
+    private config?: FileSystemResourceConfig;
```
rahulkarajgikar, line 39: incorrect optional

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,764 @@
...
+                normalizedPath === normalizedRoot
+            );
+        });
+    }
+
+    /**
+     * Check if a path is a blob storage directory that should be excluded
+     * from filesystem resource scanning to avoid conflicts with BlobResourceHandler
+     */
+    private isBlobStorageDirectory(canonicalPath: string): boolean {
```
rahulkarajgikar, line 130: this blobPatterns paradigm is very fragile and will need to be improved. will check later

packages/core/src/resources/internal-registry.ts
```diff
@@ -0,0 +1,764 @@
...
+        // Skip blob storage directories to avoid conflicts with BlobResourceHandler
+        if (this.isBlobStorageDirectory(canonical)) {
+            return;
+        }
+
+        const maxDepth = this.config?.maxDepth ?? FileSystemResourceHandler.DEFAULT_MAX_DEPTH;
+        const maxFiles = this.config?.maxFiles ?? FileSystemResourceHandler.DEFAULT_MAX_FILES;
+        const includeHidden = this.config?.includeHidden ?? false;
+        const includeExtensions =
+            this.config?.includeExtensions ?? FileSystemResourceHandler.DEFAULT_INCLUDE_EXTENSIONS;
```
rahulkarajgikar, line 308: avoid default overrides, config should already have defaults since it's been through zod

packages/core/src/resources/internal-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    async removeResourceHandler(type: string): Promise<void> {
+        if (this.handlers.has(type)) {
+            const handler = this.handlers.get(type);
+            try {
+                // Optional cleanup if supported by handler
+                if (handler && typeof (handler as any).dispose === 'function') {
+                    await (handler as any).dispose();
+                }
```
rahulkarajgikar, line 131: nonce of the handlers' support this and it's not mentioned in the interface either. can delete this try catch block entirely

packages/webui/components/types/resources.ts
```diff
@@ -0,0 +1,11 @@
+export type ResourceMetadata = {
```
rahulkarajgikar, line 1: dup: pull from core types

rahulkarajgikar (approved):
tested basic functionality, verified it works. approving and will deal with comments in a subsequent refactoring PR. 

no blocking functionality comments, mainly style/consistency violations

style/consistency related comments:
 - not following existing paradigms in APIs. idea here is to use zod for all input validations, use middleware for all error handling, which reduces most of the code and ensures we have a standard pattern everywhere. if you see if-else conditions for input payload in apis, that's a clear sign that this pattern is not being followed.
 - unnecessary use of optional arguments in every module (optional configs, optional services), leading to more if-else code which is not needed (if defined, do this, else do that). most of the optional arguments are actually always defined. so the if-else code is bloat. 
 - breaking index/types paradigms for exports- new export files like core/internal/reference are not following paradigms we have in the code base (index.ts, types.ts is the convention we are following right now) and just add bloat
 - using default fallbacks for config fields that already have defaults baked in once parsed through zod - leads to extra code for checking if the fields are defined and then applying defaults which are duplicates of what is already there in zod. these defaults would actually never be used because the defaults are always defined once passed through zod. avoid this by using `Validated` configs everywhere in internal services.  if you see code places with ``config?.field ?? DEFAULT_FIELD_VALUE```, typically in constructors, this is a violation.
 - dupe types in webUI. easy way to avoid this is to ask coding agent to review it's own code and avoid duplicates in webUI, add exports in core if necessary. 
 - not all util modules have tests - this is fine for now since we've verified it works, but adding tests for complicated utility functions early on saves a lot of manual verification time and catches breakages in the future
 - blob storage being separate from existing storage module is unintuitive

design comments:
- internal prompts provider paradigm is a bit weird, we are harcdoding the same internal prompts folder for all agents. this means coding prompt goes into image editor agent as well which is unintuitive. it's probably better to stick to agent specific prompts. we can do this by having starterPrompts(short, quick ones) and file based prompts (for longer prompts, referenced using @dexto.agent_dir in the YML), and MCP prompts (prompts from the servers)
- resource module is kind of inconsistent with other modules (3 export files, internal-registry file which defines 2 handlers, would be better off split into a handlers folder with  each handler defined)
- lack of docstrings for major classes like ResourceManager/handlers/providers makes it hard to understand the purpose/responsibilities of these classes and how they are different



will see if i can add these as guidelines/rules for codex/claude or linter rules to catch them early. if not, check these in future PRs or add custom prompts to deal with these things

PR #361, opened 2025-09-17
Update tsup config to not bundle and to fix @core alias with post build script

rahulkarajgikar:
This should improve tree-shaking in consumer dependencies

Trying this out

### Release Note

- [x] Changeset added via `pnpm changeset` (select packages + bump)
  - Packages: ...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resolution of internal imports in built output to prevent broken imports; added warnings when aliases can’t be resolved.

* **Chores**
  * Switched core to an unbundled Node-focused build and removed the browser bundle.
  * Excluded test files from build outputs.
  * Added an automated post-build step that fixes published import paths.
  * Prepared a patch release declaration for the core package.

* **Refactor**
  * Minor internal rename in CLI shutdown handling (no behavioral change).

### DIFF
--- a/.changeset/common-maps-deny.md
+++ b/.changeset/common-maps-deny.md
@@ -1,0 +1,5 @@
+---
+'@dexto/core': patch
+---
+
+Update build to not bundle
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -267,7 +267,7 @@ async function bootstrapAgentFromGlobalOpts() {
     const shutdown = async () => {
         try {
             await agent.stop();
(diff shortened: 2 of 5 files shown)


rahulkarajgikar:
@coderabbitai can you explain the difference between tsup build: false and build: true, what it implies, when to pick each, and if this PR makes sense

rahulkarajgikar:
other options to explore:

keeping bundle: true and adding new settings like splitting, tree-shake, sourcemap into tsup.config.js (refer other packages using tsup)

PR #393, opened 2025-10-11
Added new Approval System + MCP Elicitaion support for HIL

shaunak99:
## Summary by CodeRabbit

- New Features
  - Unified user approval system replacing tool confirmation, supporting approvals, cancellations, and form-based elicitation.
  - WebUI: inline approval cards with dynamic JSON Schema forms, approve/deny actions, and “remember this choice.”
  - CLI: interactive approval prompts (arrow-key confirmations) and elicitation forms.
  - New internal tool "ask_user" to request structured input from users.

- Documentation
  - Events and API docs updated to approvalRequest/approvalResponse with new payloads.

- Chores
  - Default agent config enables internalTools.ask_user.

### DIFF
--- a/packages/cli/src/api/server.ts
+++ b/packages/cli/src/api/server.ts
@@ -839,12 +839,7 @@ export async function initializeApi(
             }
             try {
                 const data = JSON.parse(messageString);
-                // Support both 'approvalResponse' and legacy 'toolConfirmationResponse'
-                if (
-                    (data.type === 'approvalResponse' ||
-                        data.type === 'toolConfirmationResponse') &&
-                    data.data
-                ) {
+                if (data.type === 'approvalResponse' && data.data) {
                     // Validate the approval response payload with Zod schema
                     const validationResult = ApprovalResponseSchema.safeParse(data.data);
                     if (!validationResult.success) {
--- a/packages/core/src/approval/manager.ts
+++ b/packages/core/src/approval/manager.ts
@@ -70,6 +70,7 @@ export class ApprovalManager {
 
     /**
      * Create appropriate approval provider based on configuration
+     * TODO: Move this into factory.ts
      */
     private createProvider(
         agentEventBus: AgentEventBus,
--- a/packages/webui/types/approval.ts
+++ b/packages/webui/types/approval.ts
@@ -4,6 +4,8 @@
  * These types represent approval events as they come over WebSocket.
  * They are derived from but separate from the core approval types,
  * as they represent the serialized WebSocket payload format.
+ *
+ * TODO: Consolidate these types from core/typed payloads later
  */
 
 import type { JSONSchema7 } from 'json-schema';


packages/core/src/approval/manager.ts
```diff
@@ -0,0 +1,218 @@
...
+
+        logger.debug(
+            `ApprovalManager initialized with mode: ${config.mode}, timeout: ${config.timeout ?? 120000}ms`
+        );
+    }
+
+    /**
+     * Create appropriate approval provider based on configuration
+     */
+    private createProvider(
```
rahulkarajgikar, line 75: nit: move this factory function to `approval/providers/` folder as factory.ts

packages/core/src/approval/schemas.ts
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Schema for approval types
+ */
+export const ApprovalTypeSchema = z.nativeEnum(ApprovalType);
+
+/**
+ * Schema for approval status
+ */
+export const ApprovalStatusSchema = z.enum(['approved', 'denied', 'cancelled']);
```
rahulkarajgikar: use approvalstatus enum which already defines this

packages/core/src/approval/types.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Custom approval types for extensibility
+     * Metadata format defined by consumer
+     */
+    CUSTOM = 'custom',
+}
+
+/**
+ * Status of an approval response
```
rahulkarajgikar: these types are duplicates of info already present in the schema


1. keep enums here, this is fine
2. for duplicate types that are also there in the schemas, use schemas as source of truth and derive types from schemas (z.input/z.output)

packages/core/src/approval/types.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Type of approval being requested
+     */
+    type: ApprovalType;
+
+    /**
+     * Optional session identifier to scope the approval
+     */
+    sessionId?: string;
```
rahulkarajgikar: probably better to be mandatory, approvals would always be scoped to a session right?

packages/core/src/approval/types.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * Status of the approval
+     */
+    status: ApprovalStatus;
+
+    /**
+     * Optional session identifier (should match request if provided)
+     */
+    sessionId?: string;
```
rahulkarajgikar: same as above, mandatory

packages/core/src/approval/manager.ts
```diff
@@ -0,0 +1,218 @@
...
+            default:
+                throw ApprovalError.invalidConfig(`Unknown approval mode: ${config.mode}`);
+        }
+    }
+
+    /**
+     * Request a generic approval
+     */
+    async requestApproval(details: ApprovalRequestDetails): Promise<ApprovalResponse> {
+        const request = EventBasedApprovalProvider.createRequest(details);
```
rahulkarajgikar: `EventBasedApprovalProvider.createRequest` - this function is independent of even based approval (mentioned actionsin comment where this function is defined)

packages/core/src/approval/providers/event-based-approval-provider.ts
```diff
@@ -0,0 +1,193 @@
...
+     * Get list of pending approval request IDs
+     */
+    getPendingApprovals(): string[] {
+        return Array.from(this.pendingApprovals.keys());
+    }
+
+    /**
+     * Helper to create an approval request with generated ID
+     */
+    static createRequest(details: ApprovalRequestDetails): ApprovalRequest {
```
rahulkarajgikar: move this method into a separate file in this same folder, since it is generic to all providers and not specific to event-based-approval provider, and approval manager calls this regardless of which provider is being used underneath

packages/core/src/mcp/manager.ts
```diff
@@ -70,13 +70,27 @@ export class MCPManager {
...
+    setApprovalManager(approvalManager: any): void {
+        this.approvalManager = approvalManager;
+        // Update all existing clients with the approval manager
+        for (const [_name, client] of this.clients.entries()) {
+            if (client instanceof MCPClient) {
+                client.setApprovalManager(approvalManager);
+            }
+        }
+    }
+
```
rahulkarajgikar, line 99: avoid using setter to set this. we know that approval manager will always be present in our services.

it's better to move this as a required argument in mcpmanager and mcpclient constructors and pass it when creating the mcp manager and mcp clients, the dependency order is easier to track, and it also removes the need for `if (!approvalManager)` checks in mcpclient

setter creates a state where the service doesn't actually know if the approvalmanager exists or not and you need defensive checks anywhere (it is similar to specifying approvalmanager as an optional service)

one argument for keeping this is if we want to expose McPManager as an independent service to end-users. (because then they don't need to know about approvalmanager)

maybe let's just add a `TODO: consider making this mandatory service and setting in constructor instead`  on the setter to revisit this later, i'm realizing it might be worth it to keep it the way it is - let's just fix the `any` type to use the ApprovalManager class and keep

packages/core/src/approval/manager.ts
```diff
@@ -0,0 +1,218 @@
...
+import type { AgentEventBus } from '../events/index.js';
+import { logger } from '../logger/index.js';
+import { ApprovalError } from './errors.js';
+
+/**
+ * Configuration for the approval manager
+ */
+export interface ApprovalManagerConfig {
+    mode: 'event-based' | 'auto-approve' | 'auto-deny';
+    timeout?: number;
```
rahulkarajgikar: timeout should be mandatory (already set from zod), and we can drop fallback default below

packages/core/src/tools/confirmation/approval-based-confirmation-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+import type { ApprovalManager } from '../../approval/manager.js';
+import { logger } from '../../logger/index.js';
+
+/**
+ * Tool confirmation provider that uses the new ApprovalManager system.
+ *
+ * This provider bridges the old tool confirmation interface with the new
+ * generalized approval system, providing backward compatibility while
+ * enabling future extensibility.
+ */
```
rahulkarajgikar: i think now with most of the event-based/conditional logic for approvals moved into the approvals module, there's not a lot of reason to have this tools/confirmation module anymore

so we could probably get rid of this folder:`tools/confirmation` and just move the 4 methods in this class into the tools manager, and update the toolsmanager to have dependency on approval manager. check with codex/claude on suggestions. this will simplify toolsmanager to just have an explicit dependency on approval manager

can do this now or add a todo, should be a small change

packages/webui/components/ToolConfirmationHandler.tsx
```diff
@@ -1,38 +1,47 @@
...
-import { AlertTriangle, Wrench, Clock } from 'lucide-react';
 import { useChatContext } from './hooks/ChatContext';
 
-interface ToolConfirmationEvent {
-    executionId: string;
-    toolName: string;
-    args: any;
-    description?: string;
-    timestamp: Date;
+export interface ApprovalEvent {
```
rahulkarajgikar: this type is already in core i think and twice in webui.

ask claude/codex to analyze all the web UI changes/CLI changes for any type/utility func duplication and import from core instead - a custom slash command for claude for this could be helpful actually because we can run every PR through it.

a similar one for `any` types could be helpful - for our webui code the LLMs are choosing to create duplicate types (sometimes even in every file) or just skipping types completely by using `any`

packages/webui/components/InlineApprovalCard.tsx
```diff
@@ -0,0 +1,289 @@
...
+                />
+                {hasError && <p className="text-xs text-red-500">{formErrors[fieldName]}</p>}
+            </div>
+        );
+    };
+
+    const handleApprove = () => {
+        if (isElicitation) {
+            // Validate form
+            const schema = (approval.metadata as any).schema;
```
rahulkarajgikar: agent: remove `any`, get the right type from core. check this file for other inconsistencies like this (any type casts, duplicate types from core)

packages/core/src/events/index.ts
```diff
@@ -199,22 +199,22 @@ export interface AgentEventMap {
...
 
-    /** Fired when tool confirmation is requested */
-    'dexto:toolConfirmationRequest': {
-        toolName: string;
-        args: Record<string, any>;
-        description?: string;
-        executionId: string;
-        timestamp: Date;
+    /** Fired when user approval is requested (generalized approval system) */
+    'dexto:approvalRequest': {
```
rahulkarajgikar, line 205: check if we need to update our docs to reflect new events

packages/core/src/utils/service-initializer.ts
```diff
@@ -82,30 +84,41 @@ export async function createAgentServices(
...
+    logger.debug('Initializing approval manager');
+    const approvalManager = new ApprovalManager(agentEventBus, {
+        mode: config.toolConfirmation.mode,
+        timeout: config.toolConfirmation.timeout,
+    });
+    logger.debug('Approval system initialized');
+
+    // 5.1 - Wire approval manager into MCP manager for elicitation support
+    mcpManager.setApprovalManager(approvalManager);
+    logger.debug('Approval manager connected to MCP manager for elicitation support');
```
rahulkarajgikar: nit: MCP manager depends on approval manager so we should create approval manager first to preserve the dependency ordering

packages/core/src/tools/internal-tools/implementations/ask-user-tool.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Internal tool for asking the user questions during agent execution
+ * Leverages the ApprovalManager to prompt the user cross-platform (CLI, WebUI)
+ */
+export function createAskUserTool(approvalManager: ApprovalManager): InternalTool {
+    return {
+        id: 'ask_user',
+        description:
+            'Collect structured input from the user through a form interface. ONLY use this tool when you need: 1) Multiple fields at once (e.g., name + email + preferences), 2) Pre-defined options/choices (use enum for dropdowns like ["small","medium","large"]), 3) Specific data types with validation (boolean for yes/no, number for quantities). DO NOT use for simple conversational questions - just ask those naturally in your response. This tool is for form-like data collection, not chat. Examples: collecting user profile info, configuration settings, or selecting from preset options.',
```
rahulkarajgikar, line 41: when would this tool be used? is this separate from elicitation requests made by the MCP servers?

packages/cli/src/api/server.ts
```diff
@@ -838,9 +839,23 @@ export async function initializeApi(
             }
             try {
                 const data = JSON.parse(messageString);
-                if (data.type === 'toolConfirmationResponse' && data.data) {
-                    // Route confirmation back via AgentEventBus and do not broadcast an error
-                    activeAgent.agentEventBus.emit('dexto:toolConfirmationResponse', data.data);
+                // Support both 'approvalResponse' and legacy 'toolConfirmationResponse'
+                if (
+                    (data.type === 'approvalResponse' ||
+                        data.type === 'toolConfirmationResponse') &&
```
rahulkarajgikar: remove legacy support

packages/core/src/approval/manager.ts
rahulkarajgikar: add moving this into factory.ts as a TODO item

packages/webui/types/approval.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * WebUI Approval Types
+ *
+ * These types represent approval events as they come over WebSocket.
+ * They are derived from but separate from the core approval types,
+ * as they represent the serialized WebSocket payload format.
+ */
+
```
rahulkarajgikar, line 10: add todo: consolidate these types from core/typed payloads later

PR #437, opened 2025-10-30
feat: spawn_agent internal tool

shaunak99:
## Release Notes

* **New Features**
  * Added sub-agent spawning capability for delegating tasks to specialized agents
  * Introduced built-in Code Review Specialist and General-Purpose Analysis agents
  * Added session type and metadata support for better organization and filtering
  * Implemented session filtering UI with support for primary and sub-agent sessions
  * Enhanced session display with sub-agent metadata (type, depth, agent identifier)

* **Improvements**
  * Improved session management with scope-based filtering and hierarchical organization
  * Added session lifecycle management for spawned sub-agents (ephemeral or persistent)


packages/core/src/session/session-manager.ts
```diff
@@ -207,7 +332,15 @@ export class SessionManager {
...
-            const session = new ChatSession({ ...this.services, sessionManager: this }, id);
+            // Note: Restored sessions use parent agent config, not custom sub-agent configs
+            // This is intentional as agentConfig is session-creation-time only
+            const session = new ChatSession(
+                { ...this.services, sessionManager: this },
+                id,
+                undefined, // agentConfig - not restored
+                existingMetadata.scopes.parentSessionId, // parentSessionId - restore for event forwarding
+                existingMetadata.scopes.agentIdentifier // agentIdentifier - restore for sub-agent metadata
+            );
```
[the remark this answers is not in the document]
shaunak99: We dont want to maintain backward compatibility can ignore, main branch does not have these scopes, its a PR specific enhancement

packages/core/src/session/session-manager.ts
```diff
@@ -159,18 +208,101 @@ export class SessionManager {
...
+                'parentSessionId is required for sub-agent sessions'
+            );
+        }
+
+        const subAgentMetadata: SubAgentMetadata = {
+            parentSessionId,
+        };
+
+        return { type: 'sub-agent', subAgentMetadata };
+    }
```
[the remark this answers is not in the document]
shaunak99: `SubAgentCoordinator.spawn()` now validates parent session existence at line 182-185 before creating the
   sub-agent session, throwing `SessionError.parentNotFound(parentSessionId)` if the parent doesn't exist
  in storage.

packages/core/src/config/error-codes.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Config-specific error codes
+ * Includes file operations, parsing, and validation errors for configuration management
+ */
+export enum ConfigErrorCode {
```
rahulkarajgikar: we should remove this config directory and loader dependencies

packages/core/src/session/chat-session.ts
```diff
@@ -134,17 +151,21 @@ export class ChatSession {
             sessionManager: import('./session-manager.js').SessionManager;
         },
         public readonly id: string,
-        logger: IDextoLogger
+        logger: IDextoLogger,
+        agentConfig?: import('../agent/schemas.js').AgentConfig
```
rahulkarajgikar: looking at this class we actually dont need an agentconfig argument or any new fields to this class i think

constructor already accepts custom `services` argument which has everything needed for dependency injection

we should pass custom services here for the necessary overrides  when creating the chatsession instead of re-creating new ones here

this way sessionmanager handles the service override/initialization/dependency injection logic and we don't need any extra initialization logic in chatsession which is a better pattern. (session handles actual execution logic, manager handles creation/orchestration logic)

packages/core/src/config/agent-reference-resolver.ts
```diff
@@ -0,0 +1,345 @@
...
+import type { AgentConfig } from '@core/agent/schemas.js';
+import { loadAgentConfig } from './loader.js';
+import { ConfigError } from './errors.js';
+import { logger } from '../logger/index.js';
+import { findDextoSourceRoot } from '@core/utils/execution-context.js';
+import { getDextoPath } from '@core/utils/path.js';
+
+/**
+ * Agent reference types for spawn_agent
+ * Currently supports: built-in names and file paths
```
rahulkarajgikar: this file (and folder) is a duplicate of same one in agent-management pkg

also do we need support for loading from file paths?

adding this undoes what i did in the last PR for decoupling core from fs/path dependencies (which are node-js only) and adds back the entire config folder which i deleted which has file dependencies

i'm thinking we can start with a simpler implementation with just inline and built-in and add file path support later if we do need it -> to support built-in agents, we can hardcode/inline built-in agent configs as objects/variables in the codebase instead of yml files to avoid file system dependencies for loading them.

if we really do need file loading support, we need to figure out a way to optionally import file path dependencies only when the sub-agent tool is being used

another way to get this to work is to let the spawn_agent tool accept agent config objects and validate them in the tool itself. this way the agent spawning sub-agent can generate config objects itself and use it for spawning agents without needing to load any files.

i faced similar problems with the filesystemservice we introduced for internal tools which is mainly used in the coding agent but adds file path dependencies to the entire core package so it was hard to refactor

the end goal here is that agents which don't depend on file system/path operations should not need to import fs/path libraries which are node-js only (unlocks cloudflare workers deployment and other serverless options which are cheaper and don't work with node-js deps)

packages/core/src/config/agent-reference-resolver.ts
```diff
@@ -0,0 +1,345 @@
...
+            error instanceof Error ? error.message : String(error)
+        );
+    }
+}
+
+/**
+ * Validate that an agent config is suitable for sub-agent spawning
+ * Enforces security constraints
+ */
+export function validateSubAgentConfig(config: AgentConfig): void {
```
rahulkarajgikar: this should also do regular validation that we do for normal agents along with these things

packages/core/src/agent/DextoAgent.ts
```diff
@@ -603,6 +619,146 @@ export class DextoAgent {
...
+
+            /** Timeout in milliseconds */
+            timeout?: number;
+
+            /** Parent session ID (auto-detected if called during session.run()) */
+            parentSessionId?: string;
+
+            /** Working directory for resolving relative agent paths */
+            workingDir?: string;
+        }
```
rahulkarajgikar: this looks like mostly a duplicate of spawn options

let's see if we can consolidate these to be the same shape and reuse (seems only diff is we combined 'agent' here)

+ previous comment to inline agent configs in the code, we would be able to remove workingDir argument also

packages/core/src/agent/DextoAgent.ts
```diff
@@ -603,6 +619,146 @@ export class DextoAgent {
...
+        const parentSessionId = options?.parentSessionId || this.currentDefaultSessionId;
+
+        this.logger.info(
+            `Handing off task to sub-agent: "${options?.description || task.substring(0, 50)}..." ` +
+                `(agent: ${typeof options?.agent === 'string' ? options.agent : 'custom config'})`
+        );
+
+        try {
+            // Spawn sub-agent
+            const handle = await this.subAgentCoordinator.spawn({
```
rahulkarajgikar, line 719: rename handle to sub-agent for clarity

packages/core/src/agent/DextoAgent.ts
```diff
@@ -603,6 +619,146 @@ export class DextoAgent {
...
+                error: error instanceof Error ? error.message : String(error),
+            };
+        }
+    }
+
+    /**
+     * Run a task with timeout.
+     * @private
+     */
+    private async runWithTimeout(
```
rahulkarajgikar: don't need a separate function for running with timeout

it can be an argument/option in sub-agent handler.run which will remove unnecessary if-else code

packages/core/src/agent/sub-agent-coordinator.ts
```diff
@@ -0,0 +1,437 @@
...
+    sessionId: string;
+    depth: number;
+    agentIdentifier: string;
+    duration: number;
+}
+
+/**
+ * Handle for interacting with a spawned sub-agent
+ */
+export class SubAgentHandle {
```
rahulkarajgikar, line 64: typo for handler? 

based on how it's being used, we could just rename this to SubAgent

packages/core/src/session/session-manager.ts
```diff
@@ -11,26 +11,70 @@ import type { StorageManager } from '../storage/index.js';
...
+ */
 export interface SessionData {
     id: string;
+
+    /**
+     * Session type - for quick filtering and identification
+     * Common types: 'primary', 'sub-agent', 'scheduled', 'task'
+     * Extensible to support custom session types from plugins/extensions
+     */
+    type: string;
```
rahulkarajgikar: we can move type into metadata

and for metadata, instead of record<string,any>

something like:
(rough)
metadata: {
 type: <some enum> constraining the allowed types mentioned above rather than string
 [key]: unknown -> basically making metadata object extensible
}
(agent pls fix syntax^)

can add more known fields into this map explicitly like subAgent, task, scheduled, etc. as we know them

packages/core/src/session/session-manager.ts
```diff
@@ -199,22 +337,40 @@ export class SessionManager {
...
         const sessionKey = `session:${id}`;
-        const existingMetadata = await this.services.storageManager
+        const existingData = await this.services.storageManager
             .getDatabase()
             .get<SessionData>(sessionKey);
-        if (existingMetadata) {
+        if (existingData) {
+            const type = this.normalizeType(existingData.type);
+
+            // Backfill legacy records that lacked type
```
rahulkarajgikar: we don't need to backfill

packages/core/src/session/session-manager.ts
```diff
@@ -199,22 +337,40 @@ export class SessionManager {
...
+            if (!existingData.type) {
+                existingData.type = type;
+                await this.services.storageManager.getDatabase().set(sessionKey, existingData);
+            }
+
             // Session exists in storage, restore it
             await this.updateSessionActivity(id);
+            // Note: Restored sessions use parent agent config, not custom sub-agent configs
+            // This is intentional as agentConfig is session-creation-time only
             const session = new ChatSession(
```
rahulkarajgikar, line 355: this is where we should inject custom services if necessary

packages/core/src/agent/DextoAgent.ts
```diff
@@ -603,6 +619,146 @@ export class DextoAgent {
...
+     * );
+     *
+     * // Use a custom agent config file
+     * const result = await agent.handoff(
+     *     "Analyze the API endpoints",
+     *     { agent: "./agents/api-analyzer.yml" }
+     * );
+     * ```
+     */
+    public async handoff(
```
rahulkarajgikar, line 666: it might make more sense to let handoff accept a DextoAgent instance

otherwise we are assuming that all handoffs will create sub-agent instances

what i'm suggesting is:

instead of `spawn_agent tool -> call handoff -> create sub-agent`

`spawn_agent tool -> creates sub-agent -> passes to agent.handoff(subagent) -> agent hands off the task to the other agent without knowing if it's a sub-agent or not`

this puts the sub-agent creation as the responsibility of the spawn-agent tool and not the dextoagent class which gives two things:
1. better separation of concerns
2. possibly allowing handoffs to other non sub-agents which may be alreaady running somewhere else

something to think about, either way is ok to start with though

PR #719, opened 2026-04-13
feat: add -w/--worktree flag for creating and managing isolated git worktrees

yanurag-dev:
### Release Note

- [ ] Changeset added via `pnpm changeset` (select packages + bump)
  - Packages: ...

## New Features
* Added `--worktree` CLI option to create and manage isolated workspaces using Git worktrees
* Automatic detection of when running inside a worktree for correct project root resolution
* New exit prompt overlay with options to keep or remove the worktree when exiting

## Documentation
* Added release notes documenting the new worktree feature and CLI flag usage


packages/cli/src/index-main.ts
```diff
@@ -553,6 +557,91 @@ program
...
+                                // Skip --worktree=value
+                            } else {
+                                childArgs.push(arg);
+                            }
+                        }
+
+                        const child = spawn(dextoPath, childArgs, {
+                            cwd: worktreePath,
+                            stdio: 'inherit',
+                            env: process.env,
```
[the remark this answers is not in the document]
yanurag-dev: This is intentional. The worktree feature creates an isolated development environment — the new process should start at the worktree root, not preserve the caller's context. This is by design.

yanurag-dev:
@coderabbit resume

packages/core/src/utils/execution-context.ts
```diff
@@ -178,6 +181,26 @@ export function findDextoProjectRoot(startPath: string = process.cwd()): string
...
+    while (currentPath !== rootPath) {
+        if (existsSync(path.join(currentPath, '.git'))) {
+            return currentPath;
+        }
+        currentPath = path.dirname(currentPath);
+    }
+
+    return null;
+}
+
```
rahulkarajgikar: i'm looking to remove path/context based utils from core and move them into agent-management

this is kind of partly in progress and the code is in a bit of a weird state (there are duplicates of these files in core and in agent-management/agent-config)

since this is a CLI concern, we should not keep this in core, and move this to agent-management/agent-config and consume in CLI index main from here. 

same idea for changes in packages/core/src/utils/path.ts, please check how CLI is consuming those

packages/agent-management/src/vcs/worktree.ts
```diff
@@ -0,0 +1,326 @@
+/**
+ * Git worktree operations for Dexto
+ * Provides utilities for creating and managing git worktrees
+ */
+
+import { execFile } from 'child_process';
+import { promisify } from 'util';
+import * as path from 'path';
```
rahulkarajgikar, line 8: similar idea here

don't want core to be too coupled to filesystem/path operations which are node-js only and makes it hard to deploy core in different serverless runtimes

we should move these files into cli/agent-management packages because core is mainly meant to be just the agent loop and orchestration around it, not file system concerns

packages/core/src/vcs/errors.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+
+import { DextoRuntimeError } from '../errors/index.js';
+import { ErrorScope, ErrorType } from '../errors/types.js';
+import { VCErrorCode } from './error-codes.js';
+
+/**
+ * VCS error factory for creating git-related errors
+ */
+export class VCSError {
```
rahulkarajgikar: this entire vcs module can move into cli or agent-management packages and out of core

packages/core/src/events/index.ts
```diff
@@ -454,6 +455,13 @@ interface AgentOwnEventMapBase {
      * Fired when the agent is fully stopped.
      */
     'agent:stopped': void;
+
+    /**
+     * Fired when /exit is run in a worktree and user should be prompted for cleanup.
+     */
+    'worktree:exit-prompt': {
+        worktreePath: string;
+    };
```
rahulkarajgikar, line 465: here this is not really a core event, but i guess there is no better way to deal with this, so it's fine for now

can you add a TODO comment: refactor event system to decouple file/CLI operations from core

rahulkarajgikar (changes requested):
code seems fine, just need to move files out of core and into the right place

can you also give some examples in the PR of how to use the new argument so i can test it?

PR #814, opened 2026-06-02
refactor: replace enums with const-list-derived unions

adarshtiwari908:
Replaces 26 TypeScript enums with const arrays and union types following the pattern from PR #686. Each enum is converted to:
- A const array as the single source of truth
- A union type derived from the array
- A named-key object for dot-access compatibility

Enums migrated:
- TodoErrorCode, AgentSpawnerErrorCode, LoggerErrorCode, TelemetryErrorCode
- JsonRpcErrorCode (numeric, uses object pattern), RuntimeErrorCode, SystemPromptErrorCode, MemoryErrorCode
- MarketplaceErrorCode, PluginErrorCode, ProcessErrorCode, SessionErrorCode
- ApprovalErrorCode, ContextErrorCode, LocalModelErrorCode, HookErrorCode
- SchedulerErrorCode, PromptErrorCode, AgentErrorCode, MCPErrorCode
- RegistryErrorCode, FileSystemErrorCode, ConfigErrorCode
- ToolErrorCode, DextoLogComponent, LLMErrorCode

Closes #689

### Release Note

- [ ] Changeset added via `pnpm changeset` (select packages + bump)
  - Packages: ...

## Summary by CodeRabbit

* **Refactor**
  * Standardized error-code representation across the codebase: enums converted to runtime constant maps + derived union types.
  * Added exported read-only lists of error codes for runtime enumeration.
  * Preserved all existing error-code values and dot-access usage — no behavioral changes for consumers.
  * Improves type safety, maintainability, and tooling compatibility; patch version bumps recorded in the changeset.


adarshtiwari908:
@rahulkarajgikar raised this PR. Please review and merge if all looks good. Thanks!

packages/core/src/llm/error-codes.ts
```diff
@@ -2,37 +2,63 @@
  * LLM-specific error codes
  * Includes configuration, validation, and runtime errors for LLM operations
  */
-export enum LLMErrorCode {
+
+export const LLM_ERROR_CODES = [
```
rahulkarajgikar: This pattern keeps the literal values in two places: the `*_ERROR_CODES` array and the named compatibility object below. 

Let's reduce drift across the migrated files by making the object the source of truth and deriving the array from `Object.values(...)`, e.g. `export type LLMErrorCode = (typeof LLMErrorCode)[keyof typeof LLMErrorCode]` and `export const LLM_ERROR_CODES = Object.values(LLMErrorCode) as readonly LLMErrorCode[]`. 

Current version works, but this removes duplication and makes the code a bit easier to maintain

Would need to fix this pattern in all the files

rahulkarajgikar (changes requested):

rahulkarajgikar:
@adarshtiwari908 i've reviewed the PR, let me know if you have any questions, thanks!

packages/core/src/llm/error-codes.ts
adarshtiwari908: @rahulkarajgikar  Got it, makes sense - will flip the pattern so the object is the source of truth and derive the array via Object.values(...). Will update all the files and push shortly.

adarshtiwari908:
Hey @rahulkarajgikar - all changes implemented. Flipped the pattern across all 26 files: object is now the single source of truth with as const, type derived via (typeof X)[keyof typeof X], and array derived via Object.values(X) as readonly X[]. Ready for re-review!

